PR #5, opened 2017-07-03
Fixing color that was not visible

jrock2004:
Changing Palette Entry #9 to a better color that would be visible. Current color was not visible due to the background color. This would fix https://github.com/system76/pop-distro/issues/33


ryanleesipes:
LGTM. @jackpot51 or @djordan2 (someone with write access, please merge).

PR #26, opened 2017-10-27
Create README.md

cassidyjames:


jackpot51:
You need to escape the underscore in Pop with a backslash

cassidyjames:
Not according to [the commonmark spec](http://spec.commonmark.org/0.28/#example-354), which GitHub uses for its markdown implementation. Its syntac highligher preview is wrong. However, to avoid ambiguity, I'll add the escape character.

jackpot51:
Thanks @cassidyjames

PR #34, opened 2018-03-27
Add language specs for SCSS/SASS

isantop:
Since we focus on developers, we should include a couple of extra language definition specs for common web-development languages.


jackpot51:
Closing because default-settings does not feel like the right place to put these

PR #36, opened 2018-04-30
Enable tap to click by default

mmstick:


jackpot51:
Done in https://github.com/pop-os/session/pull/4

PR #43, opened 2019-01-02
Try multiple key servers before failing

mmstick:
If a keyserver is offline or otherwise inaccessible, alternate key servers with our key can be used to fetch from.


brs17:
@jackpot51 , correct me if I am wrong, but I vaguely recall a conversation that we didn't feel like it was worth doing this.

mmstick:
Correct. It may be better to include the key with the package, though, so that network access isn't required to upgrade the package, in the event of an offline update.

jackpot51:
@mmstick Let's do that, include our key in the package. It can be found here: https://github.com/pop-os/pop/blob/master/scripts/.iso.asc

PR #46, opened 2019-04-01
Avoid reliance on key servers for the Pop GPG key

mmstick:


debian/pop-default-settings.postinst
```diff
@@ -15,14 +68,14 @@ case "$1" in
...
-        
+
         # Add Pop proprietary repo if it is missing.
         RELEASE="$(lsb_release -cs)"
         REPO="deb http://apt.pop-os.org/proprietary ${RELEASE} main"
         KEY="204DD8AEC33A7AFF"
         if ! grep "${REPO}" /etc/apt/sources.list > /dev/null
         then
-            apt-key adv --keyserver "keyserver.ubuntu.com" --recv-keys "${KEY}"
+            echo ${ISO_KEY} | apt-key add -
```
jackpot51: Please quote `${ISO_KEY}` here

jackpot51:
To test this, you will need to remove the deb http://apt.pop-os.org/proprietary line from /etc/apt/sources.list, or try to install this package on Ubuntu

brs17:
@mmstick let's get this into bionic and disco as well, right?

jackpot51:
@brs17 @mmstick I merged it to bionic and disco

PR #47, opened 2019-04-08
Update seahorse.desktop in disco

brs17:
In 18.*, seahorse could be found at `/usr/share/applications/seahorse.desktop`. In 19.04, it is `org.gnome.seahorse.Application.desktop`. This pr updates the name in `usr/bin/pop-app-folders` because the seahorse application was not in the appropriate "System" folder.


brs17:
`./pop/scripts/apt add seahorse` moved it into the folder! :grinning:

Issue #48, opened 2019-04-10
Automatic suspend on battery is set to 20 minutes


brs17:
The "Automatic Suspend" feature is currently enabled by default on battery power for 20 minutes. It is desired to disable this feature out of the box.

jackpot51:
@brs17 check the /org/gnome/settings-daemon/plugins/power/sleep-inactive-ac-type settings:

```
dconf read /org/gnome/settings-daemon/plugins/power/sleep-inactive-ac-type
```

It should be empty if it is set to the default.

```
gsettings get org.gnome.settings-daemon.plugins.power sleep-inactive-ac-type
```

It should return this:

```
'nothing'
```

brs17:
@jackpot51 I get exactly what you described: blank output from 

`dconf read /org/gnome/settings-daemon/plugins/power/sleep-inactive-ac-type`

and `nothing` from

`gsettings get org.gnome.settings-daemon.plugins.power sleep-inactive-ac-type`

brs17:
Well I should be looking for `sleep-inactive-battery-type` instead...update coming soon.

brs17:
So
 `dconf read /org/gnome/settings-daemon/plugins/power/sleep-inactive-battery-type`
returns blank.

`gsettings get org.gnome.settings-daemon.plugins.power sleep-inactive-ac-type`
returns 'suspend'

brs17:
I should have a pr to fix this shortly

brs17:
Fixed by https://github.com/pop-os/default-settings/pull/51, https://github.com/pop-os/default-settings/pull/50, and https://github.com/pop-os/default-settings/pull/49

PR #49, opened 2019-04-10
Disable suspend when on battery and the system is inactive

brs17:
pt1/3 for https://github.com/pop-os/default-settings/issues/48


brs17:
This works in disco. Going to create prs for cosmic and bionic.

brs17:
I've gotta rename branches...

brs17:
Maybe this will work? :woman_shrugging:

PR #52, opened 2019-04-16
Set default calendar app correctly

jackpot51:
Fixes https://github.com/pop-os/gnome-control-center/issues/20


brs17:
Didn't mean to do that

brs17:
Turns out calendar files (.ics) are just silly. I think we add this patch regardless.

jackpot51:
In some hypothetical time in the future, GNOME calendar should support directly opening ICS files

PR #53, opened 2019-04-17
Fix sourceview styles

isantop:
This PR fixes our GtkSourceView styles in Gedit by also installing them for gtksourceview-4. It also re-adds our default override to the default settings package.

@brs17 For testing this, please log out or reboot, log in, and check that the default editor theme is set to Pop-Light in Gedit preferences (under "Editor & Colors")


brs17:
@isantop Shouldn't the Pop!_Dark theme in gedit be used when the dark theme is enabled?

isantop:
@brs17 Yes, but that will need to be set by gnome-control-center. We can't set that in overrides (without patching Gedit)

PR #56, opened 2019-04-19
Fix Neofetch output to display Pop! instead of Ubuntu

mmstick:
In 19.04, `neofetch` and `lsb_release`, and possibly others, are now reading from `/usr/lib/os-release` instead of `/etc/os-release`. According to the freedesktop spec, `/etc/os-release` is supposed to take precedence over `/usr/lib/os-release`, and only fall back to `/usr/lib/os-release` if `/etc/os-release` is missing. `lsb_release` and `neofetch` seem to be exhibiting broken behavior in that regard.

This will add a symlink at `/usr/lib/os-release` so that applications using either `/etc/os-release` or `/usr/lib/os-release` will get the correct os-release.

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -2,20 +2,28 @@
 
 set -e
 
+# Make sure this is in sync with prerm
+package=pop-default-settings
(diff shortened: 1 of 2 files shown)


mmstick:
The `dpkg-divert` has been added.

jackpot51:
Are there other dpkg-diverts in other scripts

mmstick:
Yes, the `prerm` had a diversion which I've adjusted now.

jackpot51:
Thanks, I am going to do some cleanup on this and then merge it

jackpot51:
neofetch

PR #58, opened 2019-05-16
Set default vm.swappiness from 60 to 10

mmstick:
Any system with a SSD, or a sufficient amount of RAM, should avoid using swap until it is necessary. Solus is setting their default swappiness to 10, so it's likely a good idea that we do similarly here.

### Testing

- Before installing the updated package, `sysctl vm.swappiness` returns 60.
- After installing the package, running `sysctl --system` will reload the systctl parameters.
- `sysctl vm.swappiness` then should return 10.

### DIFF
--- a/debian/conffiles
+++ b/debian/conffiles
@@ -1,1 +1,2 @@
 /etc/pop-os/gdm3/custom.conf
+/etc/sysctl.d/10-pop-default-settings.conf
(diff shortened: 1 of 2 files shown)


jackpot51:
I recommend any files not using alternatives be installed with the package name. For example, etc/sysctl/10-pop-default-settings.conf

mmstick:
@jackpot51 Okay, I've renamed the file.

PR #63, opened 2019-09-10
fix: Prevent bluetooth from preventing laptops from staying suspended

mmstick:
I've been hearing that the bluetooth service causes a number of laptops, including the oryp5, to be unable to remain suspended. A script similar to this is being passed around as a workaround to the problem, so it might be useful to include this workaround by default. This will disable and stop the bluetooth service immediately pre-suspension, and then re-enable and start it post-suspension. A file is used to determine if the bluetooth service was enabled prior to suspension, so that it will only be enabled and started if it was previously-enabled

### DIFF
--- a/usr/lib/systemd/systemd-sleep/bluetooth-suspend.sh
+++ b/usr/lib/systemd/systemd-sleep/bluetooth-suspend.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+RESUME_FILE=/tmp/enable-bluetooth-on-resume
+
+if test "${1}" = "pre"; then
+  systemctl is-enabled bluetooth && touch ${RESUME_FILE}
+  systemctl disable bluetooth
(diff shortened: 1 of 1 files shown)


mmstick:
Recent Reddit thread: https://www.reddit.com/r/pop_os/comments/d22wzk/laptop_will_not_stay_in_suspend/

leviport:
I've been testing this on the oryp5 and I think I've discovered something. Fully up-to-date versions of Pop!_OS (18.04, 19.10, and 19.04) do not have the issue and remain suspended with a bluetooth device paired. I was able to replicate it by installing 19.04 from a pretty old iso and not installing updates, but as soon as I ran updates, the problem went away. I'm thinking this isn't a problem anymore, so this PR probably isn't necessary. Let's keep an eye out for additional reports of the issue though.

PR #64, opened 2019-09-10
fix: Fix CUDA functionality after resuming from suspension

mmstick:
CUDA fails to initialize after resuming from suspend with `UNKNOWN_ERROR(99)` until we reload the `nvidia_uvm` module.


mmstick:
Alternatively, we could place this in our NVIDIA driver packaging.

jackpot51:
I would prefer it in the NVIDIA driver package

PR #66, opened 2019-09-20
feat: Add LOGO field to /etc/os-release file

### LINKED ISSUE #65: Provide ICON in os-release
Not required, but would be nice.

https://www.freedesktop.org/software/systemd/man/os-release.html#LOGO=

We use this in some elementary software like the settings plug and onboarding.

mmstick:
Closes #65


jackpot51:
No potential for regression so I have merged it

PR #72, opened 2019-12-31
feat: Automatically add Flathub remote to each user

mmstick:
- Only adds flathub if flatpak is installed
- Installs the remote with `/etc/pop-os/flatpak/flathub.flatpakrepo`
- Only adds the user-based remote to accounts with a UID between the `UID_MIN` and `UID_MAX` defined in `/etc/login.defs`


isantop:
Should we consider the use case where `flatpak` is removed, a user is created, and `flatpak` is subsquently reinstalled?

mmstick:
I think the only way to handle that would be to repackage flatpak with the same post install script

isantop:
Probably a good idea then to keep that as an eventual goal, but not required for release

isantop:
One other note; does this affect new user accounts?

mmstick:
This does not affect user accounts created after a package upgrade. We would need to update the user account creation logic to add the flathub repository as well.

PR #73, opened 2020-01-04
fix: Set reasonable size for dirty bytes parameters

mmstick:
The kernel default is to buffer up to 10% of system RAM before flushing writes to the disk, which is insane. By setting a reasonable number of bytes for the `dirty_bytes` parameter, we can avoid sending the system into OOM during a large file transfer.

- [LWN article from 2013 on the issue and solution](https://lwn.net/Articles/572911/)
- [Values from this Phoronix post](https://www.phoronix.com/forums/forum/phoronix/latest-phoronix-articles/1149881-fedora-32-looking-at-using-earlyoom-by-default-to-better-deal-with-low-memory-situations?p=1149909#post1149909)


bflanagin (approved):
Settings cause no ill effects on Pop!_OS 20.04 or 20.10.

Stability is improved on low end machines with 4GB of ram.

Test run:
Created large file (40GB) using dd and passed it back and forth between drives while system was under load. 
No noticeable change in battery drain on the low end system that was tested.

jackpot51 (approved):

jacobgkau:
This was only applied to Focal, would it make sense to apply to Groovy too?

PR #83, opened 2020-04-29
4.0.4 - Migrate versioning

jackpot51:
Rebase and tag when approved

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+pop-default-settings (4.0.4) focal; urgency=medium
+
+  * Migrate versioning
+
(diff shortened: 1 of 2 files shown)


brs17:
Screenshot from 2020-04-29 17-17-53

Looks good although now I see that there's another issue with certain programs not being in the correct folders. I will file a separate issue for that.

brs17:
> Screenshot from 2020-04-29 17-17-53
> 
> Looks good although now I see that there's another issue with certain programs not being in the correct folders. I will file a separate issue for that.

New issue created here: https://github.com/pop-os/default-settings/issues/84

PR #86, opened 2020-05-04
fix: Preinstall Pop and Pop-Dark Flatpak GTK Themes

mmstick:
Adds a check on startup to install the Pop and Pop-Dark GTK themes from Flathub.

Closes https://github.com/pop-os/pop/issues/972

### DIFF
--- a/usr/bin/pop-flatpak-repos
+++ b/usr/bin/pop-flatpak-repos
@@ -1,9 +1,20 @@
 #!/bin/sh
 
-if [ -e "${HOME}/.cache/pop-flatpak-repos" ]
+CACHE_PATH="${HOME}/.cache/pop-flatpak-repos"
+
+if [ -e "${CACHE_PATH}" ]
 then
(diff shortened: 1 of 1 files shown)


jackpot51:
@brs17 it would be good to try running pop-flatpak-repos from the command line to get it's output

leviport:
```
system76@pop-os:~$ pop-flatpak-repos 
Looking for matches…
Skipping: org.gtk.Gtk3theme.Pop/x86_64/3.22 is already installed
Looking for matches…


        ID                              Branch      Op      Remote      Download
 1. [✓] org.gtk.Gtk3theme.Pop-dark      3.22        i       flathub     2.0 kB / 236.1 kB

Installation complete.
```
After running this, the cache was fixed and the new behavior is what we want. So I guess all that's left is to make sure `pop-flatpak-repos` runs after the update?

mmstick:
It should be triggered now with versioning

PR #91, opened 2020-07-01
4.0.5 - Pop!_OS style MOTD

jackpot51:
**Rebase and tag 4.0.5 when merged**

This gets rid of ads and replaces links to Ubuntu documentation with links to Pop documentation in the message of the day. You can see this if you log into a Pop machine using SSH:

```
sudo apt-get install openssh-server
ssh $USER@localhost
```


jackpot51:
@jacobgkau Added a newline after the Support line in 7a76098

PR #92, opened 2020-07-15
WIP: switch to apt.pop-os.org for Ubuntu packages

jackpot51:
This will swap the Ubuntu sources to apt.pop-os.org. Can be tested on focal which the partial mirror is syncing. This will be a draft until a full mirror is created

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -83,6 +83,15 @@ case "$1" in
             fi
         done
 
+        shopt -s nullglob
+        for file in /etc/apt/sources.list /etc/apt/sources.list.d/*.list
+        do
+            # Convert to apt.pop-os.org mirror, if using us.archive.ubuntu.com
+            sed -i 's|us.archive.ubuntu.com/ubuntu|apt.pop-os.org/ubuntu|' "${file}"
+            # Convert to apt.pop-os.org/release if using PPA
+            sed -i 's|ppa.launchpad.net/system76/pop/ubuntu|apt.pop-os.org/release|' "${file}"
+        done
+
         # Add Pop proprietary repo if it is missing.
         RELEASE="$(lsb_release -cs)"
         REPO="deb http://apt.pop-os.org/proprietary ${RELEASE} main"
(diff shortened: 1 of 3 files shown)


jackpot51:
This is a big change not in terms of code but in terms of responsibility. I'd like everyone in @pop-os/engineering and @pop-os/quality-assurance to help test it for a while

jackpot51:
Also, we need to identify packages in the PPA that were manually copied, and create repos for them

jackpot51:
Keep an eye out for packages that were installed from the PPA and are installed from Ubuntu after this change.

debian/pop-default-settings.postinst
```diff
@@ -83,6 +83,15 @@ case "$1" in
             fi
         done
 
+        shopt -s nullglob
+        for file in /etc/apt/sources.list /etc/apt/sources.list.d/*.list
+        do
+            # Convert to apt.pop-os.org mirror, if using us.archive.ubuntu.com
+            sed -i 's|us.archive.ubuntu.com/ubuntu|apt.pop-os.org/ubuntu|' "${file}"
```
ids1024, line 90: So, this won't change anything if a user has switched to another Ubuntu mirror?

If we intend for our mirror to be anything other than an exact up-to-date copy of Ubuntu's archive, this seems like it could be a problem.

jackpot51: That's correct. And we don't intend it to be any different - it is a legit Ubuntu mirror. The user is still free to change it after this.

This only changes the people with the default mirror.

isantop:
We will also need to modify things to ensure that software properties adds the correct sources. Otherwise toggling options in repoman will simply revert users back to the old repos.

isantop:
If any components or suites have been toggled/added/modified by SoftwareProperties (including via Repoman, `software-properties-gtk`, or `add-apt-repository`), then the sources will have been modified from us.archive.ubuntu.com as follows:

* For `focal`, `focal-updates`, `focal-updates`, and `focal-proposed`, the URI will point at http://archive.ubuntu.com/ubuntu
* For `focal-security`, the URI will point to http://security.ubuntu.com/ubuntu

jackpot51:
This is being done in #118 now

PR #93, opened 2020-08-19
4.0.6 - Force Firefox acceleration

jackpot51:
This improves performance and lowers battery usage for rendering tasks. Make sure to test AMD (discrete and integrated), Intel, and NVIDIA graphics.

Rebase and tag when approved


mmstick:
Tagging now

PR #96, opened 2020-09-21
fix(firefox): Unset layers accceleration and enable gfx.xrender instead

mmstick:
FIx disappearing windows and choppy refresh rate

### DIFF
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,2 +1,2 @@
-pref("layers.acceleration.force-enabled", true);
+pref("gfx.xrender.enabled", true);
 pref("widget.content.gtk-theme-override", "Pop");
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,2 +1,3 @@
 pref("gfx.xrender.enabled", true);
+pref("layout.frame_rate", 144);
 pref("widget.content.gtk-theme-override", "Pop");


jackpot51:
Note for testing: we need to test this on serw12 for sure, where the issue was reported. Make sure firefox functions with a site like youtube playing a video through a suspend/resume cycle on AMD, Intel, and NVIDIA graphics systems. Video playback is an easy way to test refresh rate, and also scrolling the page while it is playing a video.

jackpot51 (dismissed):

jacobgkau:
I'm seeing a huge reduction in performance on the serw12 with this change. It's noticeable when watching a YouTube video (playback stutters), but to quantify it, I tried this framerate test: https://www.testufo.com/framerates#count=6&background=stars&pps=960 

With layers acceleration, I'm seeing the top row run at 144fps (which makes sense, because it's a 144Hz panel.) After this change, I'm seeing the same row run at 60fps, regularly dropping to ~50fps for a second (and sometimes dipping all the way to 20-30fps after clicking around between some other tabs.)

Not having either of the options enabled results in ~22fps all the time, and gfx.xrender at least seems to prevent the scroll stuttering while not causing the transparent window issue (which I am able to reproduce when layers acceleration is turned on.) I assume we're okay with accepting a 50% performance decrease in order to mitigate the transparency bug? It's a shame not being able to watch a 30fps video smoothly on such a powerful machine.

jacobgkau:
Setting `layout.frame_rate` to `144` allows the benchmark I previously linked to run at 144fps with `gfx.xrender=true`. It also makes a noticeable difference when scrolling through text pages. The only issue is that, as I understand it, setting `layout.frame_rate=144` would waste a lot of CPU/GPU cycles on systems with regular 60Hz displays. (The default value is `-1`, which is supposed to automatically figure out the desired framerate somehow, but I've seen other reports that it incorrectly identifies Linux systems as 60Hz when they're actually higher, which seemed to be what was happening.)

The `frame_rate` option doesn't stop regular 1080p30 YouTube videos from stuttering, but after going through a lot of other settings to see if anything else might help, this appears to be the best compromise. The transparency issue gets better with WebRender, but text still disappears from tabs and the URL bar upon resume. Restarting GNOME Shell (and Mutter?) seems to bring the window contents back when it becomes distorted, but the issue also occurs in KDE Plasma/Kwin, so it definitely seems like an issue in Firefox and/or the NVIDIA driver.

mmstick:
That's a disappointment. I wonder how the newly-released Firefox performs.

jacobgkau:
Firefox 81 is performing the same way:

- With no options set, unusably slow.
- With layers.acceleration, performance is normal, but the window glitches and goes transparent after suspend/resume.
- With webrender, performance is normal, but the window glitches and goes transparent after suspend/resume. (The glitches seem to sometimes be less severe with webrender, but definitely still present.)
- With xrender, performance is lower than layers.acceleration or webrender, but still much more acceptable than not having any options set.
- layers.acceleration combined with xrender just limits the fps while still causing suspend/resume glitches.
- webrender combined with xrender doesn't seem to perform any better than just xrender.

I did check to make sure NVIDIA driver 455.23 doesn't fix this, and the transparency after suspend/resume issue is still present with that version.

I've tested this PR on:
- serw12: stops transparency issue, lowers framerate unless frame_rate is set (144Hz panel)
- gaze15: stops transparency issue, lowers framerate unless frame_rate is set (120Hz panel)
- galp4: no issues before the change, no regressions after the change.
- thelio-b1.6 w/ AMD graphics (RX580): no issues before the change, no regressions after the change.
- thelio-massive-b1.2 w/ NVIDIA graphics (RTX 2080): stops transparency issue.

After this testing, I'd suggest we do set `layout.frame_rate` to 144 in addition to turning on xrender. This allows our higher-refresh-rate systems to display at (up to) their full refresh rate, and doesn't seem to significantly impact resource usage on the systems with lower refresh rates.

leviport:
I can confirm similar behavior on the oryp6: the patch fixes disappearing windows but drops the framerate if layout.frame_rate is set back to -1. I am testing after most recent commit which sets it at 144, of course, and it looks good with that change.

jacobgkau (approved):

Darkspirit:
(Hi, I'm a Firefox user voluntarily helping with triage and regression testing.)

According to https://bugzilla.mozilla.org/show_bug.cgi?id=1673184#c8, the config of Firefox 82 on PopOS is confusingly broken.

Please don't enable unmaintained legacy gfx.xrender.enabled. It prevents users from force-enabling WebRender (the new OpenGL renderer) on their own (gfx.webrender.all=true). WebRender is required for VAAPI hardware video decoding.
gfx.xrender.enabled is disabled since Firefox 46, unmaintained and about to be removed for Fission (one process per origin like Chrome).
On Linux, Firefox still uses software rendering by default. This will [change](https://bugzilla.mozilla.org/show_bug.cgi?id=1614523).


If you wanted to enable WebRender by default for a certain configuration on your own risk, you could add whitelist entries here: https://searchfox.org/mozilla-central/rev/b757a23d312f5b154887b870447ddfeae8cfcc36/widget/GfxInfoX11.cpp#536


> With layers.acceleration, performance is normal, 

Previous OpenGL renderer (layers.acceleration.force-enabled) is [deprecated](https://bugzilla.mozilla.org/show_bug.cgi?id=1558202#c1), unmaintained and has unfixed bugs. ("Compositing: OpenGL" on about:support) Please don't enable it as well.

> but the window glitches and goes transparent after suspend/resume.

WebRender (the new OpenGL renderer) and [Google Chrome](https://twitter.com/fasterthanlime/status/1303969675953438720) are affected by the same bug (design decision) of the proprietary Nvidia driver:
https://bugzilla.mozilla.org/show_bug.cgi?id=1515253#c18
> > After RTFM at https://download.nvidia.com/XFree86/Linux-x86_64/455.28/README/powermanagement.html#SystemdConfigur74e29 and following the instructions (e.g. set NVreg_PreserveVideoMemoryAllocations=1 and sudo systemctl enable ...) I was able to restore my Firefox window successfully after resume.

The other two relevant bugs are
* https://bugzilla.mozilla.org/show_bug.cgi?id=1659848 and
* on X11 without compositing: https://bugzilla.mozilla.org/show_bug.cgi?id=1663273

(Bugs of the proprietary Nvidia driver are tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=wr-nv-linux)

> webrender combined with xrender doesn't seem to perform any better than just xrender.

They can't be combined. gfx.xrender.enabled force-disables WebRender without excuse.

jackpot51:
@Darkspirit thanks for the information, we will look into our options on Monday

PR #97, opened 2020-09-29
fix(proprietary): add the Deb822-format repository on groovy

isantop:
Since the ISO is being built with the DEB822-format sources (including the Apps repo),
the postinst script needs to be adapted to add it as well if it isn't found. Otherwise,
the ISO ends up getting build with two copies of the repo, causing APT errors in the
ISO and installed system.

It attempts to look for matches anywhere in the apt configuration, so it shouldn't add a duplicate repo if it's already configured in either the sources.list file or anywhere in /etc/apt/sources.list.d

Required for pop-os/iso#267

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -68,6 +68,15 @@ fOyJI22NfTI+3SMAsMQ8L+WVQI+58bu7+iEqoEfHCXikE8BtTbJAN4Oob1lrjfOe
 -----END PGP PUBLIC KEY BLOCK-----
 "
 
+RELEASE="$(lsb_release -cs)"
+REPO="http://apt.pop-os.org/proprietary"
+SOURCE="X-Repolib-Name: Pop_OS Applications
+Enabled: yes
+Types: deb
+URIs: ${REPO}
+Suites: ${RELEASE}
+Components: main"
+
 case "$1" in
     configure)
         for divert in "${diverts[@]}"
@@ -84,12 +93,10 @@ case "$1" in
         done
 
         # Add Pop proprietary repo if it is missing.
-        RELEASE="$(lsb_release -cs)"
-        REPO="deb http://apt.pop-os.org/proprietary ${RELEASE} main"
-        if ! grep "${REPO}" /etc/apt/sources.list > /dev/null
+        if ! grep -r "${REPO}" /etc/apt/ > /dev/null
         then
             echo "${ISO_KEY}" | apt-key add -
-            echo "${REPO}" >> /etc/apt/sources.list
+            echo "${SOURCE}" > /etc/apt/sources.list.d/pop-os-apps.sources
         fi
         ;;
 
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -93,7 +93,8 @@ case "$1" in
         done
 
         # Add Pop proprietary repo if it is missing.
-        if ! grep -r "${REPO}" /etc/apt/ > /dev/null
+        shopt -s nullglob
+        grep "${REPO}" /etc/apt/sources.list{,.d/*.list,.d/*.source}
         then
             echo "${ISO_KEY}" | apt-key add -
             echo "${SOURCE}" > /etc/apt/sources.list.d/pop-os-apps.sources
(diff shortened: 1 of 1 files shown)


jacobgkau:
Confirmed this does not affect an existing Focal system with default sources (because it would have `http://apt.pop-os.org/proprietary` in `/etc/apt/sources.list` and `/etc/apt/sources.list~` from the iso), and does successfully add the deb822-format repository on a Groovy iso that does not have that repo specified in `DISTRO_REPOS`.

This _will_ affect a Focal system in the event all references to `http://apt.pop-os.org/proprietary` are removed from `/etc/apt` and then `pop-default-settings` is reconfigured. In that case, the deb822-format repository is added to the Focal system. This shouldn't happen, but if it does, the current version of Repoman won't display it, even though it will be present and working in `apt`.

This behavior seems reasonable because if a user used Repoman to delete `http://apt.pop-os.org/proprietary`, it would still be present in `sources.list~`, so if they reach this condition, they already modified their configuration files manually and would know how to remove it again after it was re-added (and it would start showing up in Repoman again once their system's updated to groovy with Repoman supporting deb822.)

Also did an installation from a Focal iso with this branch added just to confirm nothing changed; the proprietary repo is still added in regular format and not deb822 format.

jacobgkau:
After speaking with @isantop, `sources.list~` could get updated to no longer have `http://apt.pop-os.org/proprietary` if it's removed form `sources.list` and then further changes are made, and this PR is not supposed to affect Focal systems whatsoever, so holding off until we have a `master_focal` branch that won't be affected by this PR.

debian/pop-default-settings.postinst
```diff
@@ -84,12 +93,10 @@ case "$1" in
         done
 
         # Add Pop proprietary repo if it is missing.
-        RELEASE="$(lsb_release -cs)"
-        REPO="deb http://apt.pop-os.org/proprietary ${RELEASE} main"
-        if ! grep "${REPO}" /etc/apt/sources.list > /dev/null
+        if ! grep -r "${REPO}" /etc/apt/ > /dev/null
```
jackpot51: It would probably be better to use the following as a test:

```
if ! apt-cache policy | grep "${REPO}" > /dev/null
```

This would check if the repo is currently configured, ignoring `~` and `.save` files and the like

jackpot51 (changes requested):

debian/pop-default-settings.postinst
isantop: That's a good idea! I'll set that up.

isantop: Er, wait. Actually, I think we don't want that because it would re-add it if the user decided to disable it.

jackpot51: From slack:

```
shopt -s nullglob
if ! grep "${REPO}" /etc/apt/sources.list{,.d/*.list,.d/*.source} > /dev/null
```

isantop:
Implemented as discussed, by checking the active config files instead of all files.

debian/pop-default-settings.postinst
isantop: Yep, just caught that

debian/pop-default-settings.postinst
```diff
@@ -84,12 +93,11 @@ case "$1" in
         done
 
         # Add Pop proprietary repo if it is missing.
-        RELEASE="$(lsb_release -cs)"
-        REPO="deb http://apt.pop-os.org/proprietary ${RELEASE} main"
-        if ! grep "${REPO}" /etc/apt/sources.list > /dev/null
+        shopt -s nullglob
+        if ! grep "${REPO}" /etc/apt/sources.list{,.d/*.list,.d/*.source}
```
jacobgkau: I think we should be checking for `.d/*.sources` instead of `.d/*.source`. Right now, if I change `Enabled: yes` to `Enabled: no`, the check is failing and the repo is being re-enabled every time the package is reconfigured, because the file this script creates ends in `.sources`. (`.sources` is the correct extension according to the apt [documentation](https://manpages.debian.org/stretch/apt/sources.list.5.en.html).)

jacobgkau (changes requested):

jackpot51 (dismissed):

debian/pop-default-settings.postinst
jackpot51: Good catch

isantop: Thanks for catching that, fixed!

jackpot51 (approved):

jacobgkau (approved):

PR #101, opened 2020-10-05
Convert to DEB822 sources on upgrades to Groovy

isantop:
Adds a small conversion item in the postinst that will remove Ubuntu/Pop Sources from the sources.list file and add a default system.sources file in sources.list.d

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -109,11 +109,13 @@ case "$1" in
         fi
 
         # Convert system sources from one-line format to DEB822, if present
-        if grep "ubuntu.com" /etc/apt/sources.list # Generous matching, be more specific if found
+        # Generous matching, be more specific if found
+        if grep -e "ubuntu.com" -e "apt.pop-os.org/ubuntu" /etc/apt/sources.list 
         then
             # Remove system sources from sources.list
             sed -i '/.*archive.ubuntu.com.*$/d' /etc/apt/sources.list
             sed -i '/.*security.ubuntu.com.*$/d' /etc/apt/sources.list # Needed if this was added by software-properties
+            sed -i '/.*apt.pop-os.org/ubuntu.*$/d' /etc/apt/sources.list # Convert apt.pop-os.org as well.
         fi
         # Look for existing system sources, and create if not found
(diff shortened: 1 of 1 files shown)


debian/pop-default-settings.postinst
```diff
@@ -99,8 +107,22 @@ case "$1" in
             echo "${ISO_KEY}" | apt-key add -
             echo "${SOURCE}" > /etc/apt/sources.list.d/pop-os-apps.sources
         fi
-        ;;
 
+        # Convert system sources from one-line format to DEB822, if present
+        # Generous matching, be more specific if found
+        if grep -e "ubuntu.com" -e "apt.pop-os.org/ubuntu" /etc/apt/sources.list
```
jackpot51: You can add /ubuntu to all the ubuntu.com lines just like on apt.pop-os.org/ubuntu

jackpot51 (changes requested):

debian/pop-default-settings.postinst
isantop: Updated.

jackpot51 (dismissed):
I approve. But with things like this, it is good to make very certain there aren't false positives. So @pop-os/quality-assurance should take a look

leviport (dismissed):
Seems to be working as expected after the sed fixes. Here are the contents of the resulting `sources.list` file on a fresh-ish Focal install that was upgraded to Groovy: 
```
# Ubuntu Repositories





# Pop!_OS Repositories

deb http://ppa.launchpad.net/system76/pop/ubuntu groovy main

deb http://apt.pop-os.org/proprietary groovy main
```

I also tested with a copy of my sources.list file and Jacob's sources.list file, and the Ubuntu sources were removed as expected. I can't think of any potential false positives, so I think this is good to go.

jacobgkau (approved):
One observation, this re-enables proprietary if it's been disabled in legacy format (e.g. via Repoman on 20.04). I'm considering that expected behavior because pop-upgrade already does the same thing, and this is only going to happen once, during the upgrade. (If/when pop-upgrade is updated to handle deb822 instead of wiping everything and populating the default sources.list in legacy format, this may be something to think through further.)

jackpot51 (approved):

PR #105, opened 2020-10-27
Switch to using webrender for Firefox (20.10)

jackpot51:
This may require a change to the NVIDIA driver as well to enable NVreg_PreserveVideoMemoryAllocations=1


ids1024:
Closing in favor of https://github.com/pop-os/default-settings/pull/107.

PR #106, opened 2020-11-04
firefox: Do not enable xrender, or set `layout.frame_rate`

ids1024:
Xrender appears to be deprecated and have issues.

Instead of setting this here, we can have model specific fixes in system76-driver.

See also https://github.com/pop-os/system76-driver/pull/166.

Corresponding groovy PR: https://github.com/pop-os/default-settings/pull/107


jacobgkau (approved):
See https://github.com/pop-os/default-settings/pull/107#pullrequestreview-523621426.

crawfxrd (approved):

PR #107, opened 2020-11-04
firefox: Do not enable xrender, or set `layout.frame_rate`

ids1024:
Xrender appears to be deprecated and have issues.

Instead of setting this here, we can have model specific fixes in system76-driver.

See also https://github.com/pop-os/system76-driver/pull/166.

Corresponding focal PR: https://github.com/pop-os/default-settings/pull/106

### DIFF
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,3 +1,1 @@
-pref("gfx.xrender.enabled", true);
-pref("layout.frame_rate", 144);
 pref("widget.content.gtk-theme-override", "Pop");


jacobgkau (approved):
> Xrender appears to be deprecated and have issues.

For the record, it is deprecated, but we have not observed any issues with it to my knowledge. If anything, it gives marginally better performance than the default renderer under certain circumstances.

> Instead of setting this here, we can have model specific fixes in system76-driver.

This change does limit framerate to 60fps in all circumstances on the gaze15 and oryp6, so I'd suggest setting frame_rate to 120 on gaze15 and 144 on oryp6 (but not changing the renderer on those machines) in the driver. However, it should still be okay to eliminate this global configuration immediately, since this is what the default was before the serw12 fix that's been moved to the driver, and it does work with reasonable performance and no suspend/resume issues.

jacobgkau:
I noticed that https://github.com/pop-os/system76-driver/pull/166 hasn't been merged yet. It is important that this default-settings PR is not released before that one.

ids1024:
> I noticed that pop-os/system76-driver#166 hasn't been merged yet. It is important that this default-settings PR is not released before that one.

We can make sure that is released first. I do want to make sure we're in agreement about the change to make here before releasing that, since they're interrelated. (And I guess that one's technically blocked on another engineering review.)

crawfxrd (approved):

PR #108, opened 2020-11-26
Firefox: disable recommended articles from Pocket on New Tab page by default

jacobgkau:
This PR is intended to remove the "Recommended by Pocket" section of the New Tab page in Firefox by default. (For users who enjoy the selection of content, it can be turned back on using the checkbox in the Preferences page.)

### DIFF
--- a/usr/lib/firefox/browser/defaults/preferences/pop-default-settings.js
+++ b/usr/lib/firefox/browser/defaults/preferences/pop-default-settings.js
@@ -1,0 +1,1 @@
+pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,2 +1,2 @@
 pref("widget.content.gtk-theme-override", "Pop");
-pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
+pref("browser.newtabpage.activity-stream.feeds.section.topstories", "false");
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,2 +1,1 @@
 pref("widget.content.gtk-theme-override", "Pop");
-pref("browser.newtabpage.activity-stream.feeds.section.topstories", "false");
--- a/usr/lib/firefox/defaults/pref/pop-default-settings.js
+++ b/usr/lib/firefox/defaults/pref/pop-default-settings.js
@@ -1,1 +1,2 @@
 pref("widget.content.gtk-theme-override", "Pop");
+pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);


leviport (changes requested):
This doesn't seem to be changing the setting in Firefox, even after reinstalling Firefox and pop-default-settings manually.

I'm also having second thoughts about this change. While I think Firefox is better without those recommendations, I've not heard any complaints about it from users (apart from ourselves). Until there are complaints, I don't think we need to deprive Mozilla of the ad bucks. If anyone would like to weigh in, I'd love to hear additional thoughts.

jacobgkau:
> This doesn't seem to be changing the setting in Firefox, even after reinstalling Firefox and pop-default-settings manually.

Interesting behavior. If I changed from `false` to `"false"` (with quotation marks), the preference was set as a string, and the option was turned off, but not able to be turned back on from the GUI Preferences page. After putting it back to a boolean and moving it to a different location, it's off by default and able to be turned back on. Should be ready for re-testing if we want to move forward with it.

> While I think Firefox is better without those recommendations, I've not heard any complaints about it from users (apart from ourselves).

First, I'm not sure it makes sense to wait for complaints before trying to make the experience better. (If Firefox had waited for users to complain about a _lack_ of Pocket recommendations before they added them, they might not have been added yet.)

Second, there have been at least a few complaints around the web since Firefox began integrating Pocket and its features. The complaints were just not in the context of Pop!_OS, presumably because most people understand that we don't make Firefox. (Individual complaints are also easy to address with "just turn off the checkbox," because most users aren't choosing defaults for thousands of other users, and most users probably aren't reinstalling their OS weekly or more.)

When someone opens a new tab, it's probably because they're about to either navigate to a website or perform a web search, or possibly to open a bookmark. Does it make sense to display news and magazine headlines designed to distract the user and generate clicks?

Using today's articles as an example, we have several pieces of clickbait (including one taking advantage of COVID but with no timely information, and a classic "X signs that Y is good for you"), plus at least one headline stating opinion as fact, demonstrating the editorialized nature of the recommendations:

pocket-1a

pocket-2a

Last week, the first article to pop up on a new install was "23 Signs You're Secretly a Narcissist Masquerading as a Sensitive Introvert." Being a fairly aggressive headline, this was the one that drove me to check if there was a config flag for the option that we could disable, and ironically, it's also visible in this article about turning the feature off (suggesting I'm not the only one who was bothered by it): https://www.howtogeek.com/684486/how-to-disable-recommended-by-pocket-stories-in-firefox/ 

I'm not aware of any other web browsers that attempt to pick out content for their users. My point with these examples isn't that the recommendations should be "better" or different, but that it's not the web browser's job to curate content and the new tab page is not the place for "recommendations" of any kind. Our users have work to do, and when they open the default browser on a fresh install, they don't need distractions placed by a third party.

It sounded like the engineering team was split on this, so I'm not sure if it will move forward, but that's where I'm coming from with it.

isantop:
Unless Pop_OS is willing/able to make a monetary contribution towards Mozilla to offset the lost revenue from this change, I think we shouldn't change this. The setting is as easy to turn off for users who want to do so as it would be to turn on for users who want the content on their new-tab page. And based on the lack of comments about this, I think that we should avoid making changes that financially impact a project we depend on without plans for providing that lost revenue, especially when we can reasonably presume that most of our users either use or are indifferent on this feature.

If we had replacement functionality to enhance the new tab page, I think that then this could be a good change. However, lacking that I believe the current toggle for users to opt-out of this functionality is fine. Mozilla did make this change originally based on UX research and design, and given what organization we're dealing with I'm willing to trust that they are being respectful enough with our users' data (what little, if any, they collect to provide this service). Especially given that information on disabling it is [linked directly within the new tab page itself](https://getpocket.com/firefox/new_tab_learn_more).

leviport:
It's working now. Let's leave this up for another day or two and see if more people weigh in. I'll continue to mull it over myself as well.

WatchMkr:
If it were any other app, serving ads to users, we would remove it from the default Pop!_OS applications or patch it as we have done with MOTD. I see no reason Firefox is different. I approve of this change.

leviport (approved):

ids1024 (approved):

PR #109, opened 2020-12-14
fix: Set GTK IM module to ibus

mmstick:
Part 1 of fixing the emoji ibus shortcut

### DIFF
--- a/etc/profile.d/pop-im-ibus.sh
+++ b/etc/profile.d/pop-im-ibus.sh
@@ -1,0 +1,3 @@
+export GTK_IM_MODULE="ibus"
+export QT_IM_MODULE="ibus"
+export XMODIFIERS="@im=ibus"


ids1024:
https://unix.stackexchange.com/questions/260601/understanding-setting-up-different-input-methods suggests:

```bash
export GTK_IM_MODULE="ibus"
export QT_IM_MODULE="ibus"
export XMODIFIERS="@im=ibus"
```

So I guess we might want to set those variables too?

mmstick:
Yeah that'd be fine

ids1024:
Though I see those already set that way for some reason.

jacobgkau (approved):
Together with https://github.com/pop-os/desktop/pull/57, gets the `Ctrl-Shift-E` interface for typing emojis working.

jackpot51 (approved):

PR #114, opened 2021-06-21
Add ~/.local/share to XDG_DATA_DIRS

jacobgkau:
Fixes https://github.com/pop-os/beta/issues/183 for Hirsute.

The script is based on the one Flatpak uses.


jackpot51 (approved):

jackpot51 (changes requested):
I think this is wrong. https://specifications.freedesktop.org/basedir-spec/0.6/ar01s03.html seems to indicate that XDG_DATA_DIRS needs to be searched after XDG_DATA_HOME. And if XDG_DATA_HOME is not set, assueme it is $HOME/.local/share

PR #115, opened 2021-06-29
Set up COSMIC favorites on upgrade

jackpot51:
This will add them to the start, and remove them from other places, exactly once.


jackpot51:
Closing in favor of GJS implementation by @ids1024

PR #116, opened 2021-06-29
Prepend `pop-cosmic*.desktop` to `favorite-apps` in `pop-app-folders`

ids1024:
This way upgrading Pop!_OS will add these to the dock.


jackpot51 (approved):

jacobgkau (approved):
After installing this update on a system, the user with its icons modified (that was missing the three menu icons after the upgrade) had the three menu icons added to the front of its dock. Other users still only have one set of the icons. If I remove the icons manually, they do not get added back when I re-log, reconfigure the package, or run the script again.

If a user on the beta had removed these icons from their dock before receiving this update, they will be added back once when this update is installed.

PR #118, opened 2021-08-16
Use apt.pop-os.org/release for main Pop!_OS repository

jackpot51:
We will target not using Launchpad for 21.10

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+pop-default-settings (5.1.0) impish; urgency=medium
+
+  * Use apt.pop-os.org/release for main Pop!_OS repository
+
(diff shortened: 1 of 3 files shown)


jacobgkau:
This is working to switch an Impish machine (I did have to run the provided command `sudo apt-get dist-upgrade -o Dpkg::Options::="--force-overwrite"` for updates to succeed after switching.)

However, this is placing the new URL in the same location as the old URL, in a classic/one-line format `.list` file (which still includes `ppa` in its name). Our other non-PPA repositories, including the apt.pop-os.org/proprietary repo, are using deb822. Should this repo also be switched to deb822? (The only exception I was aware of was PPAs, since Launchpad has one-line instructions for those.)

PR #120, opened 2021-11-30
feat(faces): Add pop-robot.png

mmstick:
To be used by the installer

Required by https://github.com/pop-os/installer/issues/235


jacobgkau (approved):
Places the image file in the specified location, the image can be viewed.

jackpot51 (approved):

PR #122, opened 2021-12-22
fix: Missing app icons / default application selection from nix installs

mmstick:
Useful until we have better nix packaging for Pop.

Requires logging out and logging back in after this script is added.

Can be tested by installing some applications with nix-env and then checking if the launcher shows the icons.

### DIFF
--- a/etc/profile.d/pop-nix.sh
+++ b/etc/profile.d/pop-nix.sh
@@ -1,0 +1,6 @@
+# Enables application icons to appear
+if command -v nix-env > /dev/null; then
+  DEFAULT_PROFILE=/nix/var/nix/profiles/default/share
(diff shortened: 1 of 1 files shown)


etc/profile.d/pop-nix.sh
```diff
@@ -0,0 +1,6 @@
+# Enables application icons to appear
+if command -v nix-env /dev/null; then
```
jackpot51: Don't you want to pipe into /dev/null?

mmstick: Yeah

mmstick:
Also seems to fix the ability to set a nix application as a default application.

PR #133, opened 2022-03-25
fix: Limit journald logs to 100M

mmstick:
Thoughts on lowering the default of 10% disk capacity for logs? Value can still be overridden in `/etc/systemd/journald.conf`.

### DIFF
--- a/usr/lib/systemd/journald.conf.d/pop.conf
+++ b/usr/lib/systemd/journald.conf.d/pop.conf
@@ -1,0 +1,2 @@
+[Journal]
+SystemMaxUse=1000M


jackpot51 (approved):
I think it makes sense

jacobgkau:
100M seems a little low. My work machine is at 4.1G (about 0.4% of the disk size), and a relatively fresh lab machine is already at 145M (about 0.05% of its 256GB disk).

This could impact support as well as private troubleshooting. Maybe give it at least 1GB?

jacobgkau (approved):
Confirmed this setting is working. On a test Jammy machine, generated several GB of logs, then installed the update; confirmed the oldest logs were removed and new logs are still appended, with the `/var/log/journal` directory being limited to ~1.1G.

There is no situation where this will increase the maximum log size, because the disk would have to be 10GB or less for that to be the case, and Pop!_OS will not install to a drive that small.

I did try setting the limit to a higher number (2048M) manually via `/etc/systemd/journald.conf`, and our lower limit of 1000M was still adhered to. I also tried `/etc/systemd/journald.conf.d/override.conf`, and that one worked to allow a larger journal. Looking at `man journald.conf`, it sounds like a vendor override needs to be overridden with a local override with a filename higher up in alphabetical order (overrides in `/usr` override the main `/etc` config file.) Not sure if it would be cleaner to just do this in our systemd package so we're setting the compile-time default and the main `/etc` file can still work, but I don't think doing it this way with the override would be a huge problem.

mmstick:
Overrides can also be done if you symlink a similarly-named file in `/etc/systemd/` to `/dev/null`, from what I read in the manual.

PR #137, opened 2022-04-25
fix: Add with-{alsa,jack,pulseaudio} media-session files

mmstick:
Enables Jack support out of the box. These files should be in place to enable features used by each of these frontends. Requires https://github.com/pop-os/desktop/pull/87 and https://github.com/pop-os/pipewire/pull/1


jackpot51 (approved):

n3m0-22 (approved):
JACK works out of the box now, and I have found no regressions.

PR #138, opened 2022-04-27
fix: pipewire-media-session conflict

mmstick:
This should fix upgrades that had `pipewire-media-session` installed failing to install the pop-default-settings package.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,5 +19,6 @@ Depends: ${misc:Depends},
          python3-repolib (>> 1.3.9),
          ubuntu-minimal,
          ubuntu-standard,
+Conflicts: pipewire-media-session
 Description: default settings for Pop OS
  This package contains the default settings used by Pop.


jacobgkau (approved):
`pipewire-media-session` was installed automatically in 21.10 as a dependency of `gnome-remote-desktop` (which was recommended by `pop-desktop`), so just having `pipewire-media-session` was not causing upgrades to fail. Explicitly marking it as installed did not change this.

In 22.04, `pipewire-media-session` already conflicts with `wireplumber`, which is a dependency of `pop-desktop`. `pop-default-settings` is also dependency of `pop-desktop`. I suppose this change will ensure that if `pop-default-settings` is installed prior to `wireplumber`, then `pipewire-media-session` will still be removed before `pop-default-settings` is installed (which is necessary because `pop-default-settings` actually does conflict with `pipewire-media-session` in its config files-- so this change is semantically correct as well.)

PR #139, opened 2022-04-29
Reduce default input volume to 30%

### LINKED ISSUE #400: Default mic level for laptop's built-in mic is quite a bit higher than it should be
**How did you upgrade to 22.04? (Fresh install / Upgrade)**
Clean install
 

 
Maybe gnome-control-center?

**Issue/Bug Description:**
After install, the mic input level is set way up near 80%, which is way higher than it needs to be for most laptops. I usually set mine between 20% and 30% on my oryp6 and darp5.
Screenshot from 2022-04-15 18-55-47

1. Fresh install
2. Check Settings > Sound > Input volume

**Expected behavior:**
Default input level for built-in mic should be around 20%-30%

**Other Notes:**

13r0ck:
Closes pop-os/beta#400

### DIFF
--- a/usr/bin/pop-app-folders
+++ b/usr/bin/pop-app-folders
@@ -175,7 +175,6 @@ echo "5" > "${CACHE_PATH}"
 
 fi
 
-
 if [ "${CACHE_VERSION}" -lt "6" ]
 then
 
@@ -190,5 +189,17 @@ then
 fi
 
 echo "6" > "${CACHE_PATH}"
+
+fi
+
+if [ "${CACHE_VERSION}" -lt "7" ]
+then
+
+echo "Migrating to version 7"
+
+# Default Input volume is too loud. Reduce to 30%
+pactl set-source-volume $(pactl get-default-source) 30%
+
+echo "7" > "${CACHE_PATH}"
 
 fi


usr/bin/pop-app-folders
```diff
@@ -192,3 +191,15 @@ fi
...
 
 fi
+
+if [ "${CACHE_VERSION}" -lt "7" ]
+then
+
+echo "Migrating to version 7"
+
+# Default Input volume is too loud. Reduce to 30%
+pactl set-source-volume $(pactl get-default-source) 30%
```
jackpot51, line 201: This won't help if the input device changes later. Any thoughts on where the default for all devices gets set?

It would be good to use quotes around `$(pactl get-default-source)`.

n3m0-22 (approved):
I can confirm this is working. Tested on a fresh install of 22.04 with no manual adjustments made to any sound settings.

Before:

Screenshot from 2022-04-29 13-10-05

After:

Screenshot from 2022-04-29 14-05-25

usr/bin/pop-app-folders
13r0ck: ~~Yes, I know that. From talking to Levi the issue was specifically about the internal mic on the dev one, do we want to expand this issue to all devices?~~ [see](https://github.com/pop-os/default-settings/pull/139#issuecomment-1113705769)

13r0ck:
@leviport changed his mind, he now wants the default input reduced to 30% on all input devices. So this will not work

PR #141, opened 2022-06-08
Use apt.pop-os.org for Ubuntu mirror

jackpot51:
This will switch users to from us.archive.ubuntu.com/ubuntu to apt.pop-os.org/ubuntu for the Ubuntu mirror. Needs extended testing.


mmstick (approved):
This is working from Poland, and with much faster downloads.

leviport:
So far so good on lemp9, oryp6, and dev one. I will continue to test this for the next week or two, just to make sure I don't run into trouble.

n3m0-22:
I've been testing on several machines all week with no issues so far.

n3m0-22 (approved):
This looks good.

PR #140, opened 2022-06-07
Add bluetooth suspend/resume script

jackpot51:
A simple script that just turns off bluetooth when suspending and turns it on when resuming. Could be improved by saving the device state before resume and restoring it.

PR #142, opened 2022-06-13
Bluetooth suspend fix (20.04)

jackpot51:
Same as #140 but for 20.04

### DIFF
--- a/debian/pop-default-settings.install
+++ b/debian/pop-default-settings.install
@@ -1,2 +1,3 @@
 etc
+lib
 usr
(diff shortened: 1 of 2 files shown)


mmstick (approved):

n3m0-22 (approved):
Tested on `gaze15` and `darp6` by running `sudo fwts s3 --s3-min-delay 60 --s3-max-delay 60 --s3-multiple 10`. I checked on each suspend/resume cycle that the bluetooth devices connected had reconnected before the next suspend. All tests passed. This should be good to go.

PR #144, opened 2022-06-28
bluetooth suspend fix with save rfkill state

### LINKED ISSUE #143: Bluetooth activates automatically after turning on Pc after suspension
Pop_Os! 22.04 LTS on Dell Inspiron 15 5510
 
 
Bluetooth

Nine times on ten Bluetooth activates automatically after turning on Pc after suspension

n3m0-22:
This extends the original scripts functionality by creating a temporary file before suspend if the user has turned off bluetooth. The file is removed after resume. This is to prevent the script from turning bluetooth back on after a suspend/resume cycle if it has been intentionally turned off by the user. 

May resolve https://github.com/pop-os/default-settings/issues/143

### DIFF
--- a/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
+++ b/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
@@ -2,15 +2,19 @@
 
 set -e
 
-#TODO: save rfkill state?
 case "$2" in
     suspend | hybrid-sleep)
         case "$1" in
             pre)
-				rfkill block bluetooth
+                if [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 1 ]; then
+                    > /home/$USER/blocked.bt
+                elif [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 0 ]; then
+                    rfkill block bluetooth
+                fi
                 ;;
             post)
-				rfkill unblock bluetooth
+                test ! -f /home/$USER/blocked.bt && rfkill unblock bluetooth
+                rm /home/$USER/blocked.bt
                 ;;
         esac
         ;;


XV-02 (approved):
This is working as desired.

mmstick (approved):

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
```diff
@@ -2,15 +2,19 @@
 
 set -e
 
-#TODO: save rfkill state?
 case "$2" in
     suspend | hybrid-sleep)
         case "$1" in
             pre)
-				rfkill block bluetooth
+                if [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 1 ]; then
```
pbui, line 9: What if there is more than one `bluetooth` device?  On my system I have 3 entries:

```
$ ls  /var/lib/systemd/rfkill/*bluetooth
/var/lib/systemd/rfkill/pci-0000:00:14.0-usb-0:7:1.0:bluetooth
/var/lib/systemd/rfkill/pci-0000:00:14.0-usb-0:7:bluetooth
/var/lib/systemd/rfkill/platform-dell-laptop:bluetooth
```

Therefore, the `if` statement fails because it returns too many values:

```
$ [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 0 ]
-bash: [: too many arguments
```

A possible workaround is to sort and take the highest one:

```
[ $(cat /var/lib/systemd/rfkill/*bluetooth | sort -r | head -n 1) -eq 0 ]
```

This will guarantee that you only get one value (and the highest one).

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
```diff
@@ -2,15 +2,19 @@
...
+                    > /home/$USER/blocked.bt
```
pbui, line 10: The location for this file... seems odd.  First it is not clear that `$USER` is defined within the context of systemd.  Moreover, it seems kinda hacky to place it a users home directory.   Since this is a more system-level state variable, it might be better to put it in `/run`

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
```diff
@@ -2,15 +2,19 @@
...
+                elif [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 0 ]; then
```
pbui, line 11: This has the same issue as above... perhaps you should introduced a variable:

```
BT_BLOCKED=$(cat /var/lib/systemd/rfkill/*bluetooth | sort -r | head -n 1)
case $BT_BLOCKED in
0)
rfkill block bluetooth
;;
1)
> /run/bluetooth.blocked
;;
esac
```

Using a `case` statement would also allow us to handle the case where there is no bluetooth device.

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
n3m0-22: I checked this with a second bluetooth adapter. I see why this is an issue. This seems like a good fix to me.

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
n3m0-22: This works.

lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
n3m0-22: I agree this is a better location.

jackpot51:
Please quote all shell variables!

PR #145, opened 2022-06-30
fix: bluetooth save state

pbui:
This is a follow up to #144 that handles the situation where there are
multiple bluetooth entries in /var/lib/systemd/rfkill.  The current
version would lead to a shell error and would not save the state
properly.  This new version sorts the output of the bluetooth entries
and picks the highest one.

Moreover, storing a system-level runtime file in /home/$USER seems
inappropriate.  This new version stores the bluetooth state file in
/run, which is where such things normally are stored.

I've also introduced a few variables to make the code a bit cleaner.

### DIFF
--- a/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
+++ b/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
@@ -2,19 +2,26 @@
 
 set -e
 
+BT_BLOCK_PATH=/run/bluetooth.blocked
+BT_STATES_PATH=/var/lib/systemd/rfkill/
+BT_TMP_PATH=/tmp/
+
 case "$2" in
     suspend | hybrid-sleep)
         case "$1" in
             pre)
-                if [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 1 ]; then
-                    > /home/$USER/blocked.bt
-                elif [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 0 ]; then
+                if rfkill -o ID,TYPE,SOFT | grep -q -E 'bluetooth\s+unblocked'; then
+                    cp "$BT_STATES_PATH"*bluetooth "$BT_TMP_PATH"
                     rfkill block bluetooth
+                else
+                    > "$BT_BLOCK_PATH"
                 fi
                 ;;
             post)
-                test ! -f /home/$USER/blocked.bt && rfkill unblock bluetooth
-                rm /home/$USER/blocked.bt
+                cp -f "$BT_TMP_PATH"*bluetooth "$BT_STATES_PATH" 2> /dev/null
+                [ ! -f "$BT_BLOCK_PATH" ] && rfkill unblock bluetooth
+                rm "$BT_BLOCK_PATH" 2> /dev/null
+                rm "$BT_TMP_PATH"*bluetooth 2> /dev/null
                 ;;
         esac
         ;;
--- a/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
+++ b/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
@@ -20,8 +20,8 @@ case "$2" in
             post)
                 cp -f "$BT_TMP_PATH"*bluetooth "$BT_STATES_PATH" 2> /dev/null
                 [ ! -f "$BT_BLOCK_PATH" ] && rfkill unblock bluetooth
-                rm "$BT_BLOCK_PATH" 2> /dev/null
-                rm "$BT_TMP_PATH"*bluetooth 2> /dev/null
+                rm -f "$BT_BLOCK_PATH" 2> /dev/null
+                rm -f "$BT_TMP_PATH"*bluetooth 2> /dev/null
                 ;;
         esac
         ;;


pbui:
As noted in https://github.com/pop-os/default-settings/pull/144, the current version of the script does not work if there are multiple bluetooth entries:

```
$ ls  /var/lib/systemd/rfkill/*bluetooth
/var/lib/systemd/rfkill/pci-0000:00:14.0-usb-0:7:1.0:bluetooth
/var/lib/systemd/rfkill/pci-0000:00:14.0-usb-0:7:bluetooth
/var/lib/systemd/rfkill/platform-dell-laptop:bluetooth
```

On such system, the original `if` condition fails:

```
$ [ $(cat /var/lib/systemd/rfkill/*bluetooth) -eq 0 ]
-bash: [: too many arguments
```

**Because of this, the current version of the script fails to store the state.**

This new version resolves this by sorting the bluetooth entries and picking the one with the higest value.

Moreover, storing a system-level runtime file in `/home/$USER` seems inappropriate.  Likewise it is not clear that `$USER` is defined within the context of the systemd sleep hook.  So instead, I've updated the script to store the state file in `/run` where these things typically go.

n3m0-22 (changes requested):
This is working well under most conditions, but there is a case where it will fail.

Steps to reproduce:

  1.  turn on built in bluetooth
  2. add a usb bluetooth adapter
  3. turn off bluetooth ( this will turn off both the usb and internal bluetooth )
  4. unplug usb bluetooth adapter
  5. turn on built in bluetooth

This creates a condition where `cat /var/lib/systemd/rfkill/*bluetooth` will return:
```
1
0
```

In this situation the internal bluetooth is on, but the script will still pick the highest value. In this case the highest value 1 is the now unplugged usb bluetooth adapter. This will cause the script to not run `rfkill block bluetooth` and prevent suspend since the internal bluetooth is on. 

Maybe this can be accounted for by adding a variable like the following:

`BT_DEVICE_COUNT=$(bluetoothctl list | wc -l)`

The output of this could be used to check if there is more than one bluetooth adapter in use.

jackpot51:
Please quote all shell variables

pbui:
I have updated the PR to quote all shell variables.

@n3m0-22 Would it make sense to sort the bluetooth states and select the lowest value?  That way if you have `1 0`, you would select `0` and then call `rfkill block`.  Basically if any bluetooth device is in the `0` state, then we need to block it before suspend.

Otherwise, I'm not sure how to use the `BT_DEVICE_COUNT` variable you are suggesting.

pbui:
Nevermind about selecting the lowest value... that doesn't work.

n3m0-22:
I was thinking about something like this. It could probably be done in a better way, but so far it's working for me. Of course this will only work with up to two bluetooth devices.
```
#!/bin/sh

set -e

BT_BLOCK_PATH=/run/bluetooth.blocked
BT_DEVICE_COUNT=$(bluetoothctl list | wc -l)

case "$2" in
    suspend | hybrid-sleep)
        case "$1" in
            pre)
                case "$BT_DEVICE_COUNT" in
                2)
                    BT_BLOCK_VALUE=$(cat /var/lib/systemd/rfkill/*bluetooth | sort -r | head -n 1)
                    case "$BT_BLOCK_VALUE" in
                    0)
                        rfkill block bluetooth
                        ;;
                    1)
                        > "$BT_BLOCK_PATH"
                        ;;
                    esac
                    ;;
                1)
                    BT_BLOCK_VALUE=$(cat /var/lib/systemd/rfkill/*bluetooth | sort | head -n 1)
                    case "$BT_BLOCK_VALUE" in
                    0)
                        rfkill block bluetooth
                        ;;
                    1)
                        > "$BT_BLOCK_PATH"
                        ;;
                    esac                
                esac
                ;;
            post)
                [ ! -f "$BT_BLOCK_PATH" ] && rfkill unblock bluetooth
                rm "$BT_BLOCK_PATH" 2> /dev/null
                ;;
        esac
        ;;
esac
```

n3m0-22:
I see the new code you just pushed. It's definitely a simpler solution. I'll test it now.

pbui:
Rather than checking the entries in `/var/lib/systemd/rfkill/*bluetooth`, I have changed the code to parse the output of `rfkill`.  If any of the bluetooth devices listed are in the **unblocked** state, then the script will **block** all bluetooth devices before suspending.

This should work in the normal case and in the case @n3m0-22 mentions above.

The only problem with this is that if you have multiple devices in different states, then this will block all devices, and then unblock them if any one device is unblocked before suspend.  This is not a huge deal as most people do not have multiple bluetooth devices (and not in different states), but it is a possibility.

We could save and restore the state of each bluetooth device by parsing the output of `rfkill`, but before I attempt that, I'd like to know if that is something you'd want.

n3m0-22:
I think using `rfkill -o ID,TYPE,SOFT | grep -q -E 'bluetooth\s+unblocked'` to obtain the device state is a better option since it will only show the attached devices eliminating the issue I listed above. I suppose it's possible someone could be using a bluetooth adapter to take the place of an older bluetooth device on their machine. It could cause an issue for them if they have the older bluetooth blocked and the newer running. In that case it would be good to save and restore or keep as is the state of each device between suspend/resume cycles.

n3m0-22:
The current code is working well, and solves the issues with multiple devices.

n3m0-22:
I think it would be best to preserve the states of each device for restore on resume. We could use the json output of rfkill.

n3m0-22:
This can be used to create a file with the ID and current state for each bluetooth device.
`rfkill -r -o ID,TYPE,SOFT | grep 'bluetooth' | awk '{print $1,$3}'`

n3m0-22:
This still needs work, but I thought it would be good to share to help move this along. It is mostly working, and keeps track of each bluetooth adapters state before suspend.

[pop-default-settings_bluetooth-suspend.txt](https://github.com/pop-os/default-settings/files/9023768/pop-default-settings_bluetooth-suspend.txt)

n3m0-22:
The following is working for me. It saves and restores the states of all the bluetooth adapters.

```
#!/bin/sh

set -e

BT_BLOCK_PATH=/run/bluetooth.blocked
BT_STATES_PATH=/var/lib/systemd/rfkill/
BT_TMP_PATH=/tmp/

case "$2" in
    suspend | hybrid-sleep)
        case "$1" in
            pre)
                if rfkill -o ID,TYPE,SOFT | grep -q -E 'bluetooth\s+unblocked'; then
                    cp "$BT_STATES_PATH"*bluetooth "$BT_TMP_PATH"
                    rfkill block bluetooth
                else
                    > "$BT_BLOCK_PATH"
                fi
                ;;
            post)
                cp -f "$BT_TMP_PATH"*bluetooth "$BT_STATES_PATH"
                [ ! -f "$BT_BLOCK_PATH" ] && rfkill unblock bluetooth
                rm "$BT_BLOCK_PATH" 2> /dev/null
                rm "$BT_TMP_PATH"*bluetooth
                ;;
        esac
        ;;
esac

```

pbui:
> The following is working for me. It saves and restores the states of all the bluetooth adapters.
> 

I have updated the PR to use this version of the script (with minor changes to suppress error message for `cp` and `rm`).  This works for me, but I only have one bluetooth device on hand.

n3m0-22 (approved):
This is now working.

crawfxrd (approved):

ids1024 (approved):
I was going to question if the `rfkill` output might change by system locale, but it seems like it doesn't, and is designed to be parsable when used with `--output`.

And yeah, using `/home/$USER/blocked.bt` here doesn't seem right.

This looks good assuming it works as expected.

PR #146, opened 2022-06-30
fix: bluetooth save state

linuxgnuru:
This is a follow up to #144 that handles the situation where there are
multiple bluetooth entries in /var/lib/systemd/rfkill.  The current
version would lead to a shell error and would not save the state
properly.

This new version checks if any of the bluetooth devices are in the
blocked state by parsing the output of the `rfkill` command and performs
the appropriate action.  If any bluetooth device is unblocked, then all
bluetooth devices will be blocked before suspending.

Moreover, storing a system-level runtime file in /home/$USER seems
inappropriate.  This new version stores the bluetooth state file in
/run, which is where such things normally are stored.

I've also introduced a few variables to make the code a bit cleaner.

### DIFF
--- a/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
+++ b/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
@@ -2,19 +2,21 @@
 
 set -e
 
+BT_BLOCK_PATH=/run/bluetooth.blocked
+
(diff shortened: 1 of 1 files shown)


leviport:
already open: https://github.com/pop-os/default-settings/pull/145

PR #149, opened 2022-07-12
feat: Use kyber IO scheduler for SSDs; bfq for rotational block devices

mmstick:
I've done some reading around and it seems that we're using an I/O scheduler that isn't ideal for NVME or SATA SSDs. We're defaulting to `none`, which can cause noticeable stutter and delay on the desktop when there are many processes performing I/O at the same time. Most Linux distributions are doing the same thing. It could be the cause of some of the reported issues with the system occasionally stuttering (there's reports that DRAM-less SSDs are more affected).

We may want to think about defaulting to Kyber for NVME and SATA SSDs, and BFQ for rotational drives. Kyber is a simple (< 1000 lines of code) and efficient multiqueue scheduler that's recommended in Red Hat documentation to be used by NVME/SATA SSDs. BFQ is a complex multiqueue scheduler that has a higher CPU cost but brings low latency benefits to slow rotational storage.

Tipped by some [Phoronix comments](https://www.phoronix.com/forums/forum/phoronix/latest-phoronix-articles/1334042-linux-5-19-looking-real-good-on-the-hp-dev-one-xanmod-liquorix-also-tested) talking about BFQ and None causing issues and Kyber being a good default for SSDs, I did some searching and found [this LWN article](https://lwn.net/Articles/720675/), [this Red Hat bugzilla issue](https://bugzilla.redhat.com/show_bug.cgi?id=1851783), [this Red Hat documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/monitoring_and_managing_system_status_and_performance/setting-the-disk-scheduler_monitoring-and-managing-system-status-and-performance), and [this research paper](https://www.barkhauseninstitut.org/fileadmin/user_upload/Publikationen/2019/201912_Miemietz_RTSS_K2.pdf).

From the Red Hat bugzilla issue report, there was a demand to enable mq-deadline for all SSDs because None and BFQ caused noticeable latency/stutter on the desktop for SSDs. However, MQ Deadline is focused more towards a server-oriented workflow with higher throughput at the cost of a higher latency, so I think Kyber will be better for the desktop use case. The K2 research paper also shows that Kyber has the lowest latency between BFQ, MQ Deadline, and None.

This should require testing to see how much of an impact this makes when doing a lot of I/O in the background. Perhaps playing an intensive 3D video game that streams textures and maps constantly, while also having multiple processes copying multiple files to different drives and across the network. It'd be good if we could find some people with NVME SSDs suffering from freezes that could try the change.

Note that it is working if you get an output like this:

```
$ cat /sys/block/nvme0n1/queue/scheduler
mq-deadline [kyber] bfq none
```
The change may impact max I/O throughput since that's the tradeoff made for improving desktop responsiveness.

### DIFF
--- a/etc/modules-load.d/pop-modules.conf
+++ b/etc/modules-load.d/pop-modules.conf
@@ -1,0 +1,1 @@
+kyber-iosched
--- a/lib/udev/rules.d/60-block-pop.rules
+++ b/lib/udev/rules.d/60-block-pop.rules
@@ -1,0 +1,7 @@
+# BFQ is recommended for slow storage such as rotational block devices and SD cards.
+ACTION=="add|change", SUBSYSTEM=="block", ATTR{queue/rotational}=="1", ATTR{queue/scheduler}="bfq"
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="mmcblk?", ATTR{queue/scheduler}="bfq"
+
+# Kyber is recommended for faster storage such as NVME and SATA SSDs.
+ACTION=="add|change", SUBSYSTEM=="block", ATTR{queue/rotational}=="0", KERNEL=="nvme?n?", ATTR{queue/scheduler}="kyber"
+ACTION=="add|change", SUBSYSTEM=="block", ATTR{queue/rotational}=="0", KERNEL=="sd?", ATTR{queue/scheduler}="kyber"


JacekJagosz:
I suggest mmcblk devices to have bfq assigned as well. In my testing it improves user experience of slow SD cards and devices equipped with eMMC as main storage (often chromebooks).

mmstick:
Makes sense.

Absolucy (approved):
Seems alright to me

linuxgnuru:
Would SSD-SATA drives as well as USB flash drives (which show up as /dev/sdX) i.e. SSD on the SCSI bus be affected?

mmstick:
```
ACTION=="add|change", SUBSYSTEM=="block", ATTR{queue/rotational}=="0", KERNEL=="sd?", ATTR{queue/scheduler}="kyber"
```

This udev rule will apply kyber as the scheduler for any `/dev/sdX` device that is not rotational. So that'd include SATA SSDs and USB flash drives.

mmstick:
I think bfq would be better for USB storage. Looking for an attribute to filter USB drives.

gangwerz:
Seems to be working with the internal NVMe on the galp5-1650Ti
1. Added the iosched repo
2. Upgrade without error using `sudo apt update && sudo apt upgrade`
3. Reboot and `cat /sys/block/nvme0n1/queue/scheduler` shows the `[kyber]` scheduler used on the NVMe

I tested with a USB flashdrive, and the BFQ sched is being selected. I would assume that means my flashdrive is being marked as `rotational`.

mmstick:
Looks like USB flash drives are automatically assigned `1` to their `queue/rotational` parameter, so they're getting BFQ.

mmstick:
Looks like a hardware-related flaw in a lot of flash drives.

https://bugzilla.kernel.org/show_bug.cgi?id=90761

> Currently, there is no way to distinguish if an USB that don't have VPD is a flash disk, because VPD BDC is used. USB sticks that have VPD BDC should be set as not rotational as expected. So it's a known issue, but so far there is no way to fix it.

gangwerz:
Looks like that flaw is found on all 5 my flash drives

I can't test if the USB rule will set a properly constructed flash drive will be set to kyber. But, it's consistently setting my flash drives that have `queue/rotational = 1` to BFQ.

mmstick:
Sounds fine that flash drives and portable HDDs will get BFQ by default. I'm guessing that the only devices that'd have this chip are portable SSDs.

gangwerz (approved):
Changes look good!

ahydronous:
By getting a bit clever with `ATTR{removable}=="1"`, we can separate out eMMC from SDcards:
```
  # eMMC:   "kyber" IO scheduler
  ACTION=="add", SUBSYSTEM=="block", KERNEL=="mmcblk?",  ATTR{removable}=="0" 
  # SDcard: "bfq" IO scheduler
  ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="mmcblk?", ATTR{removable}=="1"     ATTR{queue/scheduler}="bfq"
```

eMMC performance is usually much closer to SATA3 SSDs, whereas SD cards usually fare worse than even HDDs.

eMMC in the Steam Deck:
image

Samsung Pro SDXC card in the Steam Deck:
image

PR #150, opened 2022-07-22
Trackpad keyboard suspend fix

### LINKED ISSUE #399: Keyboard and touchpad not working after suspend
**Distribution (run `cat /etc/os-release`):**
 NAME="Pop!_OS"
VERSION="18.10"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Pop!_OS 18.10"
VERSION_ID="18.10"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=cosmic
UBUNTU_CODENAME=cosmic

 
 
N/A

**Issue/Bug Description:**
After a suspend, the keyboard and touchpad do not work (on the login screen) and I have to reboot to get them working again. 

Make sure suspend when the lid is closed is enabled or auto suspend is activated. Let device enter suspend mode, then wake it and try to use keyboard and mouse.

**Expected behavior:**
Keyboard and touchpad work on the login screen.

**Other Notes:**
Tested on: System76 Lemur - lemu8

Log file attached: 
[out.txt](https://github.com/pop-os/pop/files/2672565/out.txt) (note some things have been replaced with \#'s by me)

n3m0-22:
This is to fix the issue of the trackpad and keyboard not working on some laptops after resuming from suspend. It is a modification of the original bluetooth suspend fix. I have tested it on two Thinkpads and a Vaio that were having this problem. Thanks to [cslev](https://github.com/cslev) for this [comment](https://github.com/pop-os/pop/issues/399#issuecomment-1135562575).

Fixes: https://github.com/pop-os/pop/issues/399

Note: As well as normal testing, this will need to be tested on non System76 hardware that has been affected.


crawfxrd:
The last System76 laptop to use PS/2 instead of I2C HID was I think galp4 (and oryp5?).

Do any of the other affected systems use I2C HID? I hope so.

If they do use PS/2, the better solution would be to add the i8042 quirks if they work.

n3m0-22:
After further testing on my affected machines all use PS/2 not I2C. All the following were required to get the keyboard and trackpad to work consistently.

`i8042.direct i8042.dumbkbd i8042.reset i8042.nomux i8042.nopnp i8042.noloop`

Other than manually adding them to each system, how would I go about adding all the `i8042` boot options in a PR to fix all affected systems?

jacobgkau:
> Other than manually adding them to each system, how would I go about adding all the `i8042` boot options in a PR to fix all affected systems?

Historically, this was something that was applied on a per-model basis in system76-driver, although in the past, either just nomux or reset and nomux are all that have been needed: https://github.com/pop-os/system76-driver/blob/7b26eb37e358196cb9fc9963aa358b950c9614bb/system76driver/actions.py#L585 

(A list of which products used which actions can be found in the [products.py file](https://github.com/pop-os/system76-driver/blob/7b26eb37e358196cb9fc9963aa358b950c9614bb/system76driver/products.py).)

crawfxrd:
More generally, they would be added as a quirk to the kernel in `drivers/input/serio/i8042-x86ia64io.h` (See: [commit log](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/drivers/input/serio/i8042-x86ia64io.h) for the file).

n3m0-22:
@jacobgkau I had looked at that before creating this PR, and it seemed to be an unfeasible solution as it would require knowing every laptop model in the wild having this problem. @crawfxrd Wouldn't the same be true of adding quirks to `drivers/input/serio/i8042-x86ia64io.h`, or am I misunderstanding? With this I was trying to create a solution that would work for all Pop!_OS users, and to never have to address the issue again. Maybe this was the wrong way to go about it?

crawfxrd:
I assume that these units also use S3 suspend and not S0ix? Then I expect there to be a case where this happens on boot as well, which this would not fix.

This will also not work if, for some reason, psmouse is a builtin module (as one user reported).

> it seemed to be an unfeasible solution as it would require knowing every laptop model in the wild having this problem.

They're not added all at once; they're added as needed. And yes, quirks will either be for individual boards or for specific hardware if it can be detected.

Since this is not our hardware, I  would recommend the changes be added to the kernel (which could then be upstreamed to linux-input).

n3m0-22:
> I assume that these units also use S3 suspend and not S0ix

That is correct and although I have never seen it on boot I don't doubt that's a possibility. 

> This will also not work if, for some reason, psmouse is a builtin module (as one user reported).

I missed that one, but yes this would not work.

It seems this is not a good solution to the problem based on the feedback from you both, so I will close it. I appreciate the input. I'll look into adding problem machines to the kernel instead.

zm-sintra:
Hello all,
I am having this problem with Linux Lite 6.2 and a Lenovo T440s. When I suspend the computer by closing the lid, I loose the keyboard and mouse buttons (although the mouse pointer still moves with the mouse and touchpad), but if I suspend by log out menu that problem doesn't occur.
I see above that there is a solution but I don't have (yet) enough knowledge to apply it.
Could someone please describe with detailed steps how to apply this solution?
Thank you in advance for your assistance.
ZM

PR #151, opened 2022-07-28
Add custom chrony config for nts

macifell:
These custom config files for chrony do two things;

1. Add our NTS-enabled NTP servers as sources.
2. Only allow NTS to be used when an NTS enabled server is present - this overrides the default pools since we can't edit the main config file directly, but allows them to be used if our sources are commented out or removed.

For more on `authselectmode` see the following:
[https://chrony.tuxfamily.org/doc/4.2/chrony.conf.html#authselectmode](https://chrony.tuxfamily.org/doc/4.2/chrony.conf.html#authselectmode)

I considered using `mix` instead of `prefer`, but it doesn't change much except potentially increasing the accuracy slightly. In my testing it never seemed to actually incorporate any of the insecure NTP pool servers into the time calculations with either value anyway.

There is an important consideration for a computer that lacks an RTC (like a raspberry pi) - or if a system just starts out with the wrong time. It is possible the system clock will be too far out of sync to accept the server certificates when it starts up. There are a few ways around this.
  - The user can set the clock manually to an approximation of the current time and then these sources will work.
  - The following statement can be added to a config file if this is something that is expected to happen frequently, or when manual intervention is not possible:
    ```
    nocerttimecheck 1
    ```
    This will stop chrony from validating the server certificates until the clock has been initially set after booting, but it is also a less secure option - which is why I did not add it to the configuration.

More on this:
[https://chrony.tuxfamily.org/faq.html#_using_nts](https://chrony.tuxfamily.org/faq.html#_using_nts)

Some downsides to the current approach:

- These config files will be restored whenever this package is updated, which might wipe out user modifications (though only to these files, not the main config).
- There isn't a way to fall back to insecure NTP without having a user modify the config files manually - and the changes will be reverted if this package is ever updated or reinstalled (unless the main config file is edited to prevent this).
- chrony needs to be restarted (or a `chronyc` command needs to be issued) before these config files will be used on an existing install.

Ideally I would like to have a switch somewhere in the UI to disable NTS and fall back to insecure NTP (or vice versa) that would persist even if this package were reinstalled/updated - though it seems like a lot of effort and I'm not sure how useful it will really be.

However, a temporary fallback to insecure NTP is possible with the following commands:
```
sudo mv /etc/chrony/sources.d/system76-nts.sources /etc/chrony/sources.d/system76-nts.sources.disabled
sudo systemctl restart chrony
```
and to restore NTS:
```
sudo mv /etc/chrony/sources.d/system76-nts.sources.disabled /etc/chrony/sources.d/system76-nts.sources
sudo systemctl restart chrony
```

One thing I am curious about, but haven't been able to test, is how this behaves on a brand new system. I'm not sure what the clock will be set to on a new system and at what point it will get its initial update.


Absolucy (dismissed):
lgtm

macifell:
The naming of the config file has been altered to allow for override config files to be created easily.

To permanently fallback to insecure NTP:
```
echo "authselectmode ignore" | sudo tee /etc/chrony/conf.d/20-disable-nts.conf
sudo systemctl restart chrony
```

To permanently harden the system so that the certificates are always checked (this will require the clock to be manually set if the time ever gets way off):
```
echo "nocerttimecheck 0" | sudo tee /etc/chrony/conf.d/20-always-check-certs.conf
sudo systemctl restart chrony
```

jacobgkau (approved):
Tested together with https://github.com/pop-os/desktop/pull/100.

13r0ck (approved):

etc/chrony/sources.d/system76-nts.sources
```diff
@@ -0,0 +1,3 @@
+server virginia.time.system76.com iburst nts
+server ohio.time.system76.com iburst nts
+server oregon.time.system76.com iburst nts
```
kwilczynski, line 3: Very nice!

Personally, I would love to have one in Tokyo. :smile:

PR #152, opened 2022-07-29
fix(apt): Don't show phased updates in apt update output

jacobgkau:
Ubuntu's phased updates currently do not show up in the Pop!_Shop, but they do show up in `apt update` output, despite being uninstallable with `apt upgrade`/`apt full-upgrade`. This config line makes them no longer show as available updates in `apt update` output, and prevents them from showing up as "held back" in `apt upgrade`/`apt full-upgrade` output (`apt policy` will still show if a package has a phased update in progress.)

The alternative to this would be `APT::Get::Always-Include-Phased-Updates "true";` and `Update-Manager::Always-Include-Phased-Updates "true";` to install the updates instead of hiding them, but that doesn't seem to affect the Pop!_Shop (and I can't find any separate config line for PackageKit), so it would result in inconsistent behavior between the Pop!_Shop and apt.

### DIFF
--- a/etc/apt/apt.conf.d/76pop-default-settings
+++ b/etc/apt/apt.conf.d/76pop-default-settings
@@ -1,0 +1,1 @@
+APT::Get::Never-Include-Phased-Updates "true";


jacobgkau:
Regarding the filename, all other files in this directory currently take the format `##filename` or `##longer-file-name`, with the highest default being number 70 (`70debconf`). Setting this to 76 makes it last but still leaves room for others above it in case users want to customize their own config.

mmstick (approved):

n3m0-22 (approved):
This looks good. I am no longer seeing the phased updates in the output of `apt update`. I set a watch on `apt-update` and waited till new packages were available. I did this three times installing updates through `pop-shop`, `sudo apt upgrade`, and `sudo apt full-upgrade`. Everything seem to be working correctly.

Before:

before

After:

after

PR #153, opened 2022-08-04
fix: Opt out of phasing updates by including them in updates

mmstick:
Previous fix doesn't seem to work, perhaps because different dependencies are assigned a different phased rate, and then newly-unphased packages fail to upgrade without their phased counterparts being ready to install.


13r0ck (approved):

leviport (approved):

leviport:
Whoops, forgot to add my message.

Before patch:
Screenshot from 2022-08-12 07-48-14

After: (fewer updates because I had to remove `master` to get the patch installed)
Screenshot from 2022-08-12 07-51-23

PR #154, opened 2022-08-08
Set suspend settings to 30 minutes for new installs

13r0ck:
Because of Title 20 compliance for computer manufacturers new
Pop!_OS installs running system76 machines require 30 minute
auto-suspend.


mmstick (approved):

XV-02 (approved):
New Pop installs on System76 hardware should now suspend after 30 minutes by default, on AC or Battery. Approved.

PR #156, opened 2022-08-18
Move suspend settings to pop-os/desktop

13r0ck:
It was decided that suspend should be enabled by default for all
users, not just systems manufactured by system76. As such, the
setting has been moved to pop-os/desktop
Commit: cb3460d394702ea22b283f3bec6923052aeea5ae

Requires: pop-os/desktop#101

### DIFF
--- a/usr/bin/pop-app-folders
+++ b/usr/bin/pop-app-folders
@@ -190,17 +190,12 @@ then
 fi
 
 # Title 20 Compliance requires a display off setting of 5 minutes, and suspend of 30 minutes.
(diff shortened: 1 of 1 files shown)


crawfxrd (dismissed):

jacobgkau (dismissed):
Tested with https://github.com/pop-os/desktop/pull/101.

13r0ck:
It was decided by carl that systems shouldn't be radioactively changed, so this PR has been modified

jacobgkau (approved):
This no longer affects an existing user account. It still affects a new user account on an existing system.

crawfxrd (approved):

PR #157, opened 2022-08-23
Add missing newline to 76pop-default-settings

cu:
Before:

```
$ cat /etc/apt/apt.conf.d/76pop-default-settings 
APT::Get::Always-Include-Phased-Updates "true";$
```

After:

```
$ cat etc/apt/apt.conf.d/76pop-default-settings 
APT::Get::Always-Include-Phased-Updates "true";
$ 
```


jacobgkau (approved):
The necessity of this is questionable, but I can see it being annoying not having the newline in a config file when trying to troubleshoot/investigate a system.

The file had a newline at the end before https://github.com/pop-os/default-settings/pull/153.

13r0ck (approved):

cu:
Thanks for merging this.

Standard Unix text manipulation tools work on lines of text where a line of text is defined as zero or more characters terminated by a newline. apt may be tolerant of the missing newline, but not many other tools are.

PR #160, opened 2022-09-20
Add new NTS servers

macifell:
A couple of new time servers with NTS support have been created outside of the US. This updates the chrony configuration to use them.


13r0ck (approved):

jacobgkau (approved):
Confirmed this adds the two new sources to the config file. Tested both of the new sources by commenting out the rest; both new servers were able to be selected by Chrony and were able to correct the system clock.

PR #161, opened 2022-09-27
Add sleep before rfkill unblock bluetooth

### LINKED ISSUE #158: Bluetooth off after waking from suspend 22.04
With bluetooth on, I enable suspend.  After waking from suspend bluetooth is off, I need to turn it back on.  I see the same issue was fixed for 20.04 now it's happening in 22.04 #147.  I'm on a Dell XPS 13 7390.

jravetch:
Fixes #158

### DIFF
--- a/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
+++ b/lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
@@ -19,7 +19,7 @@ case "$2" in
                 ;;
             post)
                 cp -f "$BT_TMP_PATH"*bluetooth "$BT_STATES_PATH" 2> /dev/null
-                [ ! -f "$BT_BLOCK_PATH" ] && rfkill unblock bluetooth
+                [ ! -f "$BT_BLOCK_PATH" ] && sleep 1 && rfkill unblock bluetooth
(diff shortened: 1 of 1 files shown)


mmstick (approved):

n3m0-22 (commented):
This is passing testing on `gaze15` `oryp4` and `darp6`, but I was not able to recreate the original issue. I also tested on a framework with deep sleep enabled. I have seen no regressions. This will need further testing on other hardware.

jravetch:
The issue was found in a Dell XPS 13 7390 and XPS 15 9500 both running 22.04.

n3m0-22:
Yes, I am currently looking into getting it tested on Dell hardware as well, but it also needs to be tested and not cause any regression on other hardware including System76. I was just reporting my current findings. Thank you for this contribution.

leviport:
I was able to get access to an XPS 13 9310 for testing. It appears that suspend and resume always results in bluetooth being reactivated as expected without this patch, but adding this patch doesn't cause any regressions on this laptop. If it fixes it on the 7390 and the 9500, and it doesn't cause any regressions, then I think this is good to go. @n3m0-22 feel free to approve if you're happy with it.

n3m0-22 (approved):

PR #162, opened 2022-09-27
update repoman mirror reset

### LINKED ISSUE #100: `Reset Mirrors to Defaults` is out of date (resets to Ubuntu instead of apt.pop-os.org)
Following https://github.com/pop-os/default-settings/pull/141, the default apt source should be `apt.pop-os.org/ubuntu` instead of `us.archive.ubuntu.com/ubuntu`. The `Reset Mirrors to Default` button currently still resets to the old default.

n3m0-22:
This is to update the default mirror reset in Repoman from `http://us.archive.ubuntu.com/ubuntu/` to `http://apt.pop-os.org/ubuntu`.

Fixes: https://github.com/pop-os/repoman/issues/100

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -79,7 +79,7 @@ URIs: ${PROPRIETARY_REPO}
 Suites: ${RELEASE}
 Components: main"
 
-SYS_REPO="http://us.archive.ubuntu.com/ubuntu/"
+SYS_REPO="http://apt.pop-os.org/ubuntu"
 SYS_SOURCE="X-Repolib-Name: Pop_OS System Sources
 Enabled: yes
 Types: deb deb-src


mmstick (approved):

jacobgkau (approved):
The current default source for systems upgraded from pre-mirror times includes a trailing slash, since the transition code did not remove the trailing slash: https://github.com/pop-os/default-settings/pull/162/files 

However, new installs do not have a trailing slash (which I pointed out when that PR was created): https://github.com/pop-os/default-settings/pull/141/files 

Since this PR is setting the default to not have a trailing slash, this means the reset button still shows up on systems where the source was upgraded from pre-transition times. This is still an improvement over the current behavior of resetting to the incorrect Ubuntu mirror (which then requires re-transitioning next time default-settings is reconfigured.) The only way I can think of to get around this would be both updating the transition code to remove the trailing slash, and adding a new transition _just_ to remove the trailing slash from current installations that have already been transitioned. Not sure if that's worth doing.

Approving this since it's an improvement over current behavior, especially for installs that have happened since the actual default source was changed.

PR #166, opened 2023-02-14
chore: remove Ubuntu Pro apt advertisements

mmstick:
Resolves https://www.reddit.com/r/pop_os/comments/111vajn/ubuntu_pro_advertisements_in_pop_os/

This file is included by `ubuntu-advantage-tools`, which is a required dependency of `ubuntu-minimal`.

### DIFF
--- a/debian/pop-default-settings.postinst
+++ b/debian/pop-default-settings.postinst
@@ -90,6 +90,9 @@ X-Repolib-Default-Mirror: ${SYS_REPO}"
 
 case "$1" in
     configure)
+        # Remove Ubuntu Pro advertisement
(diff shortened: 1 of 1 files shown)


n3m0-22 (approved):
I have not been able to reproduce this issue, but I can confirm this PR is removing `20apt-esm-hook.conf`. I have seen no regressions caused by this.

jackpot51 (approved):

jacobgkau:
On one of my test systems, I had to manually run `sudo dpkg-reconfigure pop-default-settings` for this to take effect. Is the `case "$1" in configure)` block supposed to get run on every update to the package, or only on installs or when other conditions are met?

PR #167, opened 2023-03-06
fix: Opt out of phased updates Focal

n3m0-22:
Phased updates messages have started showing up in Focal.

```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
#
# "Phased updates" allow careful rollout of deb package changes at scale.
# You may see these as packages that have been "kept back". Learn more at
# https://ubuntu.com/server/docs/about-apt-upgrade-and-phased-updates
#
The following packages have been kept back:
  fwupd-signed
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
```

This is an addition of the original fix for Jammy.
https://github.com/pop-os/default-settings/pull/153
https://github.com/pop-os/default-settings/pull/166

This works for fresh installs. Otherwise `sudo dpkg-reconfigure pop-default-settings` needs to be run after adding this PR.

It will still suffer from the issue pointed out by @jacobgkau https://github.com/pop-os/default-settings/pull/166#issuecomment-1478390144, but it will at least put Jammy and Focal on equal footing in regards to this issue.


leviport:
I don't think it worked:

ahoneybun:
Wouldn't it be like this?:

APT::Get::Always-Include-Phased-Updates "false";

leviport:
@ahoneybun did you see https://github.com/pop-os/default-settings/pull/153 ?

jackpot51 (dismissed):

leviport:
I might need to wait a bit longer for the cron job to run, but I still got the phased updates message after installing this update and rebooting:


I'd also like an engineering opinion on whether this is the kind of fix we'd be happy with.

jackpot51:
I expected this to work because it matches the jammy change. I would want the same thing on focal, once it works.

n3m0-22:
My reasoning for going beyond what was done for Jammy is that in Focal while `APT::Get::Always-Include-Phased-Updates "true";` does seem to opt out of phased updates it does not suppress the message from Ubuntu when running `apt upgrade`.

My concern with this is that users will still be seeing a message about phased updates with a link to Ubuntu server docs https://ubuntu.com/server/docs/about-apt-upgrade-and-phased-updates.

I do see a difference in the version of `apt` that may be part of the problem (i.e., Jammy `apt` is `2.4.8`, Focal `apt` is `2.0.6`).

I also noticed that in `/etc/apt/apt.conf.d` on Jammy there is no file `20apt-esm-hook.conf`, but there is on Focal. If this file is deleted or it's contents commented out on Focal the message does go away. Maybe removing this file on Focal is a more acceptable solution.  

If this message showing is not of concern I can remove the second commit, and it should still opt out the user.

13r0ck:
> but `ubuntu-advantage-tools` is now providing

Man we really need to improve our metapackaing to remove that package

n3m0-22:
> > but `ubuntu-advantage-tools` is now providing
> 
> Man we really need to improve our metapackaing to remove that package

If removing the package completely is ok I can confirm from my testing that solves the issue as well.

13r0ck (changes requested):
Rather than running everytime the system boots, this should just be in https://github.com/pop-os/default-settings/blob/master_jammy/usr/bin/pop-app-folders

And run `pro config unset apt_news` once

n3m0-22:
> Rather than running everytime the system boots, this should just be in https://github.com/pop-os/default-settings/blob/master_jammy/usr/bin/pop-app-folders
> 
> And run `pro config unset apt_news` once

I had thought about doing that, but I was worried an update to `ubuntu-advantage-tools` might reset the change. Although I suppose an update could also make this fix invalid anyway.

n3m0-22:
On a fresh install of Pop 20.04, with or without the changes in this PR, I'm no longer seeing any messages about phased updates. Checking `pro config show` I'm seeing `apt_news` set to `true`. This PR may no longer be necessary, but that will need to be double checked.

13r0ck (changes requested):
Sorry for the delay. This should be wrapped in the check for cache version, i.e.:
```bash
if [ "${CACHE_VERSION}" -lt "1" ]
```
So it only runs once. I know you said that you arn't seeing the issue anymore, but it won't hurt to add this encase it comes back. After this change we should merge this.

usr/bin/pop-app-folders
```diff
@@ -96,3 +96,17 @@ esac
 echo "1" > "${CACHE_PATH}"
 
 fi
+
+if [ "${CACHE_VERSION}" -lt "1" ]
```
13r0ck: Still not quite there. Look at the app-folders file on jammy. The cache version needs to be iterated after each check, and the if statement needs to increase by each time as well

13r0ck (changes requested):

13r0ck (dismissed):
Looks good!

n3m0-22:
I'm closing this for now as it is not working. I will reopen if the issues presents itself again.

jacobgkau (approved):
Before installing:
Screenshot from 2023-03-23 13-58-42
After installing:
Screenshot from 2023-03-23 13-59-26
Regarding https://github.com/pop-os/default-settings/pull/166#issuecomment-1478390144, in my Focal test environment, I did see the `leaving diversion of ...` messages when doing `apt upgrade` that would indicate the configure script ran.

jackpot51 (approved):

PR #169, opened 2023-04-05
Remove old dependencies because of pop-os/metapackage reorg

13r0ck:
requires: https://github.com/pop-os/desktop/pull/113

Will require another branch to push to focal

PR #170, opened 2023-04-07
[jammy] chore: configure apt esm hook with diversion

mmstick:


jackpot51 (commented):
I would prefer diverting to a new, empty file. This is in case a user wants to edit this file.

n3m0-22 (changes requested):
This is creating two new files:
 - `20apt-esm-hook.conf`
 - `20apt-esm-hook.conf.diverted`

After installing running `sudo apt update` while still working now shows the following:

```
All packages are up to date.
N: Ignoring file '20apt-esm-hook.conf.diverted' in directory '/etc/apt/apt.conf.d/' as it has an invalid filename extension
N: Ignoring '20apt-esm-hook.conf' in directory '/etc/apt/apt.conf.d/' as it is not a regular file
```

If the files are manually deleted and `pop-default-setting` is reinstalled only the file `20apt-esm-hook.conf` is created.

mmstick:
Sounds like we can't use diversions for this if apt will ignore them because they're symbolic links.

mmstick:
https://github.com/pop-os/desktop/pull/113

This would be a better solution since it allows us to eliminate the `ubuntu-advantage-tools` dependency that installs this file.

jackpot51 (approved):

jacobgkau:
Obsoleted by https://github.com/pop-os/default-settings/pull/169 / https://github.com/pop-os/desktop/pull/113.

PR #173, opened 2023-08-10
Set ideal MGLRU setting for desktop responsiveness

mmstick:
This sets `/sys/kernel/mm/lru_gen/min_ttl_ms` to `1000` on startup, which ensures that any memory pages accessed within the last second will remain in memory, and avoid getting swapped or reclaimed. The default setting is `0`, which disables the feature.

Based on https://aur.archlinux.org/packages/mglru-desktop-settings

I've been tipped that this can noticeably improve YouTube or PC game responsiveness on a system that's out of memory, or close to it.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -16,13 +16,19 @@ Depends: ${misc:Depends},
          gnupg,
          plymouth-theme-pop-basic,
          python3-repolib (>> 1.3.9),
-Recommends: pop-default-settings-zram
+Recommends: pop-default-settings-lru, pop-default-settings-zram
 Conflicts: pipewire-media-session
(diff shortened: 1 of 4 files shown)


jacobgkau:
I started testing this on a gaze18 today, but haven't observed too much of a difference. While filling up the RAM with `stress-ng --vm-bytes $(awk '/MemAvailable/{printf "%d\n", $2 * 1.1;}' < /proc/meminfo)k --vm-keep -m 1`, a 1080p YouTube video already played back smoothly without the change. It did freeze up for a second when opening another window on top of it, but it still does that after the change.

PR #174, opened 2023-11-06
Enable ufw by default

### LINKED ISSUE #874: Set status of ufw firewall to 'active' by default
Issue/Bug Description:
Pop!_OS comes with the ufw firewall installed, but inactive. There is no GUI application to manage the firewall, nor any indication to the user that there is a firewall installed or that it is disabled. I would like to suggest that, similar to other distributions like Fedora, that Pop!_OS have the firewall set to 'active' rather than 'inactive' by default.

Expected behavior:
The ufw firewall should ideally be activated by default in a new installation, rather than needing to be activated manually in the terminal.

image

leviport:
I believe this should enable ufw by default on Pop.

Closes https://github.com/pop-os/pop/issues/874

### DIFF
--- a/etc/ufw/ufw.conf
+++ b/etc/ufw/ufw.conf
@@ -1,0 +1,10 @@
+# /etc/ufw/ufw.conf
+#
+
+# Set to yes to start on boot. If setting this remotely, be sure to add a rule
+# to allow your remote connection before starting ufw. Eg: 'ufw allow 22/tcp'
+ENABLED=yes
+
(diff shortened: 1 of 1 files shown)


leviport:
I can't tell if I just can't test this, or if it just doesn't work

leviport:
I built an ISO with this change and it didn't enable ufw by default. There's something missing somewhere.

leviport:
Since this isn't working, I'll close and delete the branch. We can revisit it some other time if we want ufw enabled by default.

PR #175, opened 2024-11-05
fix(apt): add Signed-By field if missing

mmstick:
Adds a Signed-By field to our apt sources, as required by 24.04

Closes https://github.com/pop-os/cosmic-epoch/issues/516


leviport (approved):
Makes the "missing signed-by" messages disappear on 24.04 when doing apt updates.

jackpot51 (approved):

PR #178, opened 2026-02-09
feat: automatically set wireless reg domain

mmstick:
Automatically calls `iw reg set ${country_code}` on startup and also when the timezone changes. This will improve wifi signal in some countries by enabling more channels, frequencies and/or higher power where they are available.

Based on https://invent.kde.org/kde-linux/kde-linux/-/merge_requests/367

Should be copied to the resolute branch


garrettjwilke (approved):
changed multiple timezones (Africa, Americas, Asia, Australia)...rebooted and checked:
`iw reg get`

correctly adjusts on each boot

jackpot51 (approved):

Conan-Kudo:
Heh, this is a surprise. 😅

PR #180, opened 2026-02-23
fix: enable udisks to mount NTFS partitions with ntfs3

mmstick:
It is required to define a udev rule to force udisks to use the new `ntfs` driver over `ntfs3` so that the user needn't install the less maintained `ntfs-3g` FUSE driver (which unsafely ignores the dirty bit flag in the file system). UDISKS_FILE_SYSTEMSHARED=0 is the default behavior for other file systems, which auto-mounts to `/media/{{volume-name}}` by default.

### DIFF
--- a/debian/pop-default-settings.install
+++ b/debian/pop-default-settings.install
@@ -19,6 +19,7 @@ etc/xdg/autostart/pop-app-folders.desktop
 etc/xdg/autostart/pop-flatpak-repos.desktop
 lib/systemd/system-sleep/pop-default-settings_bluetooth-suspend
 lib/udev/rules.d/60-block-pop.rules
+lib/udev/rules.d/60-ntfs3.rules
 lib/udev/rules.d/85-iw-regulatory.rules
 usr/bin/pop-app-folders
 usr/bin/pop-cosmic-favorites
--- a/lib/udev/rules.d/60-ntfs3.rules
+++ b/lib/udev/rules.d/60-ntfs3.rules
@@ -1,0 +1,2 @@
+# Enables udisks to mount NTFS parititons with the ntfs3 module by default.
+SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_TYPE}="ntfs3" , ENV{UDISKS_FILESYSTEM_SHARED}="0"


garrettjwilke:
installed this fix, removed `ntfs-3g` and rebooted

drive mounts, can write files and folders

if you incorrectly unmount the disk, it will not mount again with:

image

attempting to fix:
```
sudo ntfsfix /dev/sda1
Mounting volume... OK
Processing of $MFT and $MFTMirr completed successfully.
Checking the alternate boot sector... OK
NTFS volume version is 3.1.
NTFS partition /dev/sda1 was processed successfully.
```

attempting to mount gives the same error...

i assume this NTFS driver works as intended, but the issues with NTFS on linux are still difficult :)

mmstick:
Perhaps it'd be helpful to automatically run `ntfsfix -b -d` on NTFS file systems before mounting.

wash2 (dismissed):

jklgrasso:
I ran into the same issue as Garrett: https://github.com/pop-os/default-settings/pull/180#issuecomment-3947409811. If we're fine with that for the time being to resolve the mounting issue, I can approve this. Just want to confirm first

jklgrasso:
On Linux 7.1.5, without `ntfs-3g`, I can mount disks and create/delete/copy/move/etc. files on an existing NTFS filesystem. But I cannot create new filesystems. Is that expected?

I'm still seeing dirty NTFS filesystems upon improper removal and have no resolution options without `ntfs-3g`. Is there another package I should/need to install (other than `ntfs-3g`), or will we continue using `ntfs-3g` alongside the  "Resurrection" driver so this won't be an issue? As far as I know, `ntfsfix` doesn't exist without `ntfs-3g`.

mmstick:
If we backport `ntfsprogs-plus`, this provides `mkfs.ntfs` for the new `ntfs` kernel driver. `ntfsfix` is still required to clear the dirty bit flag though, so it still needs to be installed for GNOME Disks to use it when using the disk repair option. We could potentially have cosmic-files run the `ntfsfix -d` command when mounting a NTFS partition with the dirty bit flag set.

jklgrasso:
That sounds good to me

PR #183, opened 2026-05-07
fix: remove pop-im-ibus profile

mmstick:
Not needed with COSMIC and will interfere with fcitx

### DIFF
--- a/debian/pop-default-settings.install
+++ b/debian/pop-default-settings.install
@@ -12,7 +12,6 @@ etc/pop-os/lsb-release
 etc/pop-os/os-release
 etc/pop-os/update-motd.d/10-help-text
 etc/pop-os/update-motd.d/50-motd-news
-etc/profile.d/pop-im-ibus.sh
 etc/sysctl.d/10-pop-default-settings.conf
 etc/update-manager/release-upgrades.d/pop-default-settings.cfg
 etc/xdg/autostart/pop-app-folders.desktop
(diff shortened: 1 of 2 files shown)


jackpot51 (dismissed):

wash2 (dismissed):

jacobgkau (changes requested):
Failed to build:

```
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
   dh_auto_test
   create-stamp debian/debhelper-build-stamp
 fakeroot debian/rules binary
dh binary
   dh_testroot
   dh_prep
   dh_auto_install
   dh_install
dh_install: warning: Cannot find (any matches for) "etc/profile.d/pop-im-ibus.sh" (tried in ., debian/tmp)

dh_install: warning: pop-default-settings missing files: etc/profile.d/pop-im-ibus.sh
dh_install: error: missing files, aborting
make: *** [debian/rules:8: binary] Error 255
dpkg-buildpackage: error: fakeroot debian/rules binary subprocess returned exit status 2
```

I think we need to remove the file from the list at https://github.com/pop-os/default-settings/blob/b272e92e09073dc1e4dce01344790e696bbbaa5f/debian/pop-default-settings.install#L15, in addition to just removing the file itself.

jacobgkau (approved):
This removes the obsolete and unneeded file from `/etc/prfile.d/pop-im-ibus.sh` that set the environment variables (which are no longer recommended for use in Wayland) to IBus (which is not the preferred IM for COSMIC).

The variables are still being set in `/etc/X11/Xsession.d/70im-config_launch` from the `im-config` package, which is a recommends of fcitx5 (even though fcitx5 shows a warning if these variables are set recommending that they be unset) and is also a recommends of ibus (which is installed by default in Pop!_OS despite not being a dependency of anything in particular, from what I can see). That's another piece we'll need to figure out how to fix elsewhere.

jackpot51 (approved):

